PR #1, opened 2019-04-07
Sync timestamps

agricolab:
As requested, i moved the pull request to the new repo structure according to https://github.com/sccn/xdf/pull/28
I had to manually copy the code, as the history is broken, and hope i did not miss anythings. Tests did run fine. 
image


agricolab:
On a side note, i added the feature before the ordering/recasting currently discussed in https://github.com/sccn/xdf/pull/42
This implementation should therefore survive, if there is a decision made for the breaking change.

cboulay:
Can you make a trivial change and commit it? I'm hoping to trigger a CI build and that it will run the test suite. Maybe add a newline at the end of test_limit_on_synced_stamps.py.

agricolab:
Seems to work!

agricolab:
The test in CI gave a warning: ```##[warning]invalid escape sequence \R```. I also get this, when i run pytest the first time after a change, from the second on, this is gone. The line 230 that causes this warning is the final ```"""``` from the docstring of ```load_xdf```. Haven't found a way to get rid of this warning. Might be a bug in pytest?

agricolab:
I guess the next action item is to rebase this commit on the changes in master.

cbrnr:
What's the status of this PR? I'm not even sure what this adds - is this implementing resampling of all streams to a common sampling frequency? @agricolab maybe you could update this PR with a short description in your top comment.

agricolab:
Updated the top comment with a direct quote from the original pull request. 

Regarding your question, yes it syncs streams by interpolation/shifting so that samples match. Regarding the status, this is relatively old now, and does therefore not (yet) account for the recent feature of subselection of streams from an xdf-file.

cbrnr:
Nice, this feature is important. I can rebase if you want.

agricolab:
That'd be great, i am currently quite busy with work, and would probably find the opportunity to work on this not before january.

cbrnr:
Alright, I rebased and fixed some PEP8-related stuff. Maybe I can come back to this before January.

cbrnr:
@agricolab I assume you don't have time to finalize this PR? In addition to rebasing, I think what's currently missing is support for stream selection (which has been added to `main`). Also, it would be nice if the frequency that is used for resampling could be defined manually. By default, it could still be the highest sampling frequency among all (selected) streams, but this is not always practical if there's e.g. an audio stream, which could also be downsampled.

agricolab:
I can make time in the next few weeks. Considering it's more been apparently five years in the making it really is time to resolve this 😅

cbrnr:
😄 This would be really great! If this is available in PyXDF, downstream projects don't have to roll their own implementations.

Speaking of which, we've been discussing the problem of resampling here: https://github.com/cbrnr/mnelab/issues/385

Specifically, we were thinking about a solution when there are long gaps between otherwise relatively regularly spaced samples. @DominiqueMakowski implemented linear interpolation (using pandas), which is certainly better than assuming that samples are regularly spaced, but I am not sure if interpolating such long periods is the best solution. I think that replacing these segments with NaNs would be a better option.

WDYT? Is this covered by your changes, and if so, how do you deal with large gaps?

agricolab:
I updated everything to integrate with the most recent version. What is not implemented yet is 
- [ ] Handle long gaps between otherwise relatively regularly spaced samples
- [ ] Frequency that is used for resampling can be defined manually

DominiqueMakowski:
Small note:

>  Handle long gaps between otherwise relatively regularly spaced samples

I would not only do a special treatment when long gaps are detected and otherwise assume that is regularly sampled. In fact, I would never assume that it is (from experience, it's often not, and a few milliseconds are vital in some applications). + LSL returns the timestamp of all samples so it seems like a shame not to use it and have a loss of precision

cbrnr:
This PR is about creating regularly sampled time series via resampling. So I think large gaps need to be treated somehow in order to avoid interpolating a large gap and instead insert e.g. NaNs.

cbrnr:
@agricolab I was wondering if this PR can also deal with just a single stream, i.e. resample that stream to a regular sampling frequency (and dealing with larger gaps). Or is this already possible with the current release?

agricolab:
With use_samplingrate, one should be able to resample a single stream

agricolab:
Regarding the large gaps in the recording, i am not sure how this will interfere or replicate stuff that's already being handled by `jitter...` and `clock_reset...` and `winsor_threshold`

agricolab:
I understand your request, but i try not to fall into a feature creep that'll result in a rewrite of pyxdf

cbrnr:
You absolutely don't have to do this if you don't want to. I was just wondering that from a user perspective, the most useful representation of a single regularly-sampled stream would be a resampled version that's based on all timestamps. This makes sure that the sampling frequency is indeed e.g. 500Hz, as opposed to 500.03231Hz as it's currently the case if you just take the time stamps as is. [np.interp()](https://numpy.org/doc/stable/reference/generated/numpy.interp.html) does exactly that. And an additional step would then be to handle larger gaps instead of interpolating, this can be done as a post-processing step.

agricolab:
Currently, regardless of whether one uses the fastest stream for syncing (i.e. it's effective srate) or sets it manually, the time_stamps are created newly to be equidistant. See https://github.com/agricolab/xdf-Python/blob/sync_timestamps/pyxdf/pyxdf.py#L985-L998

There was a reason why i did it so complicated for the highest effective srate, but I can't recall right now after so many years. Maybe I was just stupid...

I agree that how to handle large gaps is worthy of more discussion. I didn't consider it an issue because of the other routines already there which do normalize the timestamps. I'll think a little bit  about how and whether we can integrate it easily.

agricolab:
Currently, regardless of whether one uses the fastest stream for syncing (i.e. it's effective srate) or sets it manually, the time_stamps are created newly to be equidistant. See https://github.com/agricolab/xdf-Python/blob/sync_timestamps/pyxdf/pyxdf.py#L985-L998

There was a reason why i did it so complicated for the highest effective srate, but I can't recall right now after so many years. Maybe I was just stupid...

I agree that how to handle large gaps is worthy of more discussion. I didn't consider it an issue because of the other routines already there which do normalize the timestamps. I'll think a little bit  about how and whether we can integrate it easily.

agricolab:
Ok, I started implementing.

Seggments will be detected by the _jitter... Hlhelper function and I'll propagate information about start and stop to the _sync functions.

One issue is that there might be different segments for different streams, i.e. we can't simply use one stream with whom to sync the rest. I'll accommodate for that.

Additionally limiting to overlapping timestamps will get a different behavior. It had returned all overlapping samples from earliest to latest, but now will only return overlapping segments available in all streams.

cbrnr:
> One issue is that there might be different segments for different streams, i.e. we can't simply use one stream with whom to sync the rest. I'll accommodate for that. It'll be a little bit tricky because the time between a stream's segments is not in sync with the streams sampling rate.

I'm trying to think about this in simpler terms when you only have *one* stream. Or in other words, you handle each stream separately by resampling (interpolating) to evenly spaced samples, even between segments. If you do that for every stream independently, you end up with samples at the exact same time points, meaning that they should be synchronized. Regarding periods between segments, these could be filled with NaN as a post-processing step.

Or am I simplifying too much here?

agricolab:
Consider a stream with a Fs of 10 Hz. One segment might start at 1.1s and go to 2.1s, the second segment starts at 2.75 and go to 4.65. The start and stop of each segment is not aligned to the effective sampling rate within each segment.
Which segment should be shifted and how... That's the question.

cbrnr:
If you interpolate to 10Hz and start e.g. at second 1.1, then you have regularly spaced samples for the entire time period, including segments and periods in between.

Here's a little example to illustrate what I mean:

```python
import matplotlib.pyplot as plt
import numpy as np

# x ... original time values
# y ... original signal values
x = np.linspace(0, 2 * np.pi, 10)
x = x[[0, 3, 4, 6, 7, 8, 9]]
y = np.sin(x)


def interp(xvals, x, y, limit=None):
    yinterp = np.interp(xvals, x, y)
    if limit is None:
        return yinterp
    for i, interval in enumerate(np.diff(x) > limit):
        if interval:  # difference > limit, so fill with np.nan
            yinterp = np.where((xvals < x[i + 1]) & (xvals > x[i]), np.nan, yinterp)
    return yinterp


# xvals ... new (resampled) time values
# yinterp ... new (resampled) signal values
xvals = np.linspace(0, 2 * np.pi, 50)

# limit=2 means that gaps > 2 (in the original units of x) will not be interpolated, but
# filled with np.nan
yinterp = interp(xvals, x, y, limit=2)

plt.plot(x, y, 'o')
plt.plot(xvals, yinterp, '-x')
plt.show()
```

Figure_1

agricolab:
The behavior of `_jitter_removal`  which is currently used by load_xdf to detect breaks in the timestamps is to linearize the timestamps within each segment independent from all other segments 
https://github.com/agricolab/xdf-Python/blob/sync_timestamps/pyxdf/pyxdf.py#L710-L718

That means each segment can have a different effective srate, and segments do not share a common time basis, i.e. samples are not aligned on a common srate. The streams field `effective_srate` will be a duration-weighted average of all segments https://github.com/agricolab/xdf-Python/blob/sync_timestamps/pyxdf/pyxdf.py#L720C1-L728C76

For a multi-segment stream (which can always be the case), that means we 

a) start at the earliest available sample, end at the latest available sample, and resync all segments to the average effective srate (or the srate given by the argument) -> all segments will have the same srate and are aligned
b) resample all segments to the average effective srate, but keep them unaligned
c)  resample all segments to the average effective srate, and align them to the segment with the longest duration

Optionally, we can fill gaps with NaN.

Now,  because the effective srate averaged across segments will be unequal to the srate within each segment (at least due to floating point accuracy) all segments will need to be resampled anyways to their new srate. That means c) doesn't make much sense. 

If we use the stream with the highest effective srate, one could argue to leave that stream (and its segments) untouched and simply sync all other streams with the time-stamps as given by the fastest stream. That makes sense, as it maintains the original signal fidelity in `.time_series` by not interpolating. This was the original idea of this PR (while still being naive to the segmentation issue). If the fastest stream is segmented, we will likely cause loss of data for the other streams, as we will just jump the gaps between the segments of the fastest stream. This is not optimal. It could be circumvented by filling the gaps with NaNs, but then there is the issue with srate to use (either smoothen the change in srate and time basis between segments which would not affect signal fidelity but cause havoc for most signal processing routines as they can't assume a stationary srate, or sync everything to a common basis which is simple and straightforward, but results in loss of fidelity).

There is now the option to say that if the user want's to sync with the fastest stream, and this stream has only a single segment, we use that streams time_stamps and leave its time_series untouched. If there are multiple segments, we sync everything to a common basis from earliest to latest sample and fill the fastest strteams gaps with NaNs.  I don't like that because it would result in different behavior depending on the data and can not be controlled by the user.

So, the simplest default behavior would be let the user decide whether to use the fastest stream or support their own desired srate. Then, all streams (including the  fastest) will be resampled to that srate and span from the earliest sample of all streams  to the lastest sample of all streams, and gaps not supported by data from segments will be set to NaN.

cbrnr:
This sounds good to me! I understand your argument that you are reluctant to resample a stream even though it is sometimes not necessary. However, resampling (if done correctly) should not affect the fidelity of the signals. Just because you are not using the exact samples that dropped out of the amplifier does not mean that this is less reliable (or wrong in any other sense).

As a user, I basically almost always would like to have my data in a 2D array with some constant sampling frequency. So are you saying you would use the original (effective) sampling frequency and just fill gaps with NaN for the fastest stream to avoid resampling?

Another option related to this issue could be to let the user decide if they want (1) the original data (which is what they get now, each stream separately with their samples and associated time values) or (2) if they want a resampled version (i.e., a 2D NumPy array). What you are suggesting is something in between, right? The original values for the fastest stream, and resampled values for the others. I'm not sure this is worth the added complexity.

DominiqueMakowski:
I would strongly push back against resampling - or doing something to the signal & timestamps - by default (actually, if anything this PR should be non-breaking) + I would always return the rawest data (raw samples with raw timestamps so that people can postprocess if they want).

My argument against resampling by default is that it is never a trivial process, even with linear interpolation. Problems of signal aliasing can appear and strongly distort the signal.  

I strongly agree with @cbrnr to let the user decide if they want the original data or a resampled version (In which case I think that resampling to the highest existing frequency is not good enough, you need at least to double the rate in order to avoid signal aliasing)

agricolab:
Thanks for the comment. To clarify, the while idea of this PR is that a user want's to align the data from one or more streams to another stream, or resample all to a common time basis. This always requires interpolation/resampling.

If will never be default behavior, and only be performed if the user specifies the argument `sync_timestamps` to a non-False string defining the interpolation method (and or defines the sampling rate via `use_samplingrate`

cbrnr:
So yes, we all agree that the current behavior (no resampling, just spit out timestamp/value pairs for each stream independently) will remain the default.

When using `use_samplingrate` (or maybe call it `resample`), what is the proposed plan? If `resample=True`, the highest sampling frequency will be used (no need for twice the highest frequency for the reason @agricolab mentioned). If `resample=f`, `f` will be used. The output in this case will be a single 2D NumPy array, where all streams share a common sampling frequency and time basis, and any gaps are filled with NaNs.

BTW, all of this interacts with `select_streams`, right?

Do you agree?

agricolab:
Ok, my plan was not to return a 2D array, but simply the same streams, headers tuple as before. The reason is I don't want to break what load_xdf returns.

But we could do this as an optional argument and pass a target array to be filled.

agricolab:
Re select streams... I'll check. Syncing etc is done just before load_xdf returns so it should only be aware of selected streams

cbrnr:
> Ok, my plan was not to return a 2D array, but simply the same streams, headers tuple as before. The reason is I don't want to break what load_xdf returns.

The `header` part (i.e. the tuple element 1) is just the generic header of the XDF file, so basically useless. All streams are contained in element 0. So you could keep returning this 2-tuple, but element 0 could be a NumPy array in this case.

Or did you want to keep the nested list and default dict structure?

agricolab:
> As a user, I basically almost always would like to have my data in a 2D array with some constant sampling frequency.

I fully agree, but i would prefer not to break the interface of load_xdf. An alternative to passing a array to be filled as reference argument is subclassing list, e.g.

```
class XDFList (list):
    _is_synced = False
    def as_2d_array(self):
        ... magic ....
       if _is_synced:
          return 2darray if possible
       else:
          raise Exception("Not synced")
```

cbrnr:
I don't think this change is breaking the `load_xdf()` interface. By default (i.e. `resample=False`), nothing changes. Only when `resample=True` would the function return a different first tuple element (a NumPy array), and I think that's perfectly fine.

agricolab:
I am still at odds. When the function returns a numpy array, it is difficult to impossible to recover the information without loading the whole file with load_xdf again. Worse because channel format does not need to be the same between streams, the dtype of the array would be a python object which makes any further computations on the array costly and forfeits the utility of using numpy in the first place.

On the other hand, it is trivial to concatenate an array out of the list of defaultdicts.

cbrnr:
You are right, a list of defaultdicts containing synchronized NumPy arrays is the better option then!

agricolab:
Great, i'll iron out some issues and will keep you updated once it's ready for testing

chkothe:
Hey folks, someone drew my attention to this. It's an interesting undertaking to have load_xdf optionally "pretend" that one's lab instrumentation happened to generate one big array of data with NaNs for gaps. 

However, I'd argue that the feature would have to be offered with a huge asterisk, which is that it will introduce some pretty massive spectral aliasing in every stream that was either upsampled or downsampled. The TL;DR is that the spectral content will most likely look like something from gallery below this single sweep (the original):
before

Basically,  chances are that we'd hand our users something like this, depending on the type of interp-style method used:





Credit where credit is due -- these are all taken from a website that I had been pointing colleagues to for almost 15 years, https://src.infinitewave.ca/ (the gallery is much longer and it's frankly shocking, there are like 25 ways you can shoot yourself in the foot and blow your leg off when resampling). Although apparently the makers of the software compared in there have since replaced their earlier resampling attempts, so now you have to skip over relatively many examples that look like the clean chirp above (still recommended, you can use the up/down arrow on the combobox).

Since we're all having a sigproc background, I'll keep it very short (I saw it was mentioned and then quickly dismissed):
* when we upsample, then unless the interpolated content (i.e., the straight line or the quadratic function) was band-limited to the original nyquist frequency of that stream (which it isn't), the result would contain various spectral content above that original frequency, and most likely it will also include several copies of the signal below that new nyquist frequency due to the nature of those interpolant functions.
* when we downsample, then unless we had band-limited the original signal to the future nyquist frequency, it would at the very least be reflected about that frequency (in spectrum), but in addition we'll get more copies and spectral content depending on the chosen method.

The only approach that should give us a clean spectrum would be to use a dedicated resampling function like eg in MATLAB's dsp toolbox, or something like libsamplerate or libfsrc (we could also roll our own FIR-type or windowed sinc interpolator to not require users to have scipy, in which case they could just call resample themselves). However, while these interpolators preserve the spectrum, in good Heisenberg manner, they will be non-causal in time, meaning that, if there was a big spike (eg artifact) in the signal, we hand our users a dataset that has a symmetric ripple some time before, and also some time after that (time depending on the desired quality of the spectrum). Also any such interpolator has to invent a continuation of the signal at the edges (zero, flat, line, or mirrored in some way, since a band-limited signal doesn't just stop like a brick wall). Resample offers some 5 options IIRC. This wouldn't make a huge difference in practice, _except_ if there are gaps in the signal, because information would be transported across the (arbitrarily long) gap by the resampling kernel if that wasn't accounted for, i.e., across an experiment break or such.

So I think we'd have to make the tough choice of either giving our users something that's very problematic if they do any kind of spectral analysis of their data (and we can't assume that it's all just for some Master's thesis where maybe it doesn't matter), or that's quite problematic if they do any kind of causal arguments in their analysis (e.g., any offline analysis that claims to be faithful to real-time use, or connectivity / Granger causality analysis). Perhaps unsurprisingly, that's why load_xdf currently doesn't offer to "simply" resample everything to the same rate (a precursor to LSL at SCCN actually did that by default, fell flat on its face, and presumably left an imprint in some publications between 2005 and 2010). But to state the obvious, of course if a user feels the temptation to resample everything to the same rate then chances are they will simply do one of the above themselves, though hopefully using their MATLAB's or scipy's resample function, and hopefully they don't have gaps in the data (unfortunately with more devices moving to Bluetooth, that becomes ever more likely due to packet loss), but they may do something much worse. So if the XDF project shipped tooling that helps them put all their lab instrumentation to use while following best practices (and informing users about any pitfalls), the world would be better off than it is today. But it'd mean potentially a lot more work on this, and also the argument would be more like a dictionary with options for at least the method, the edge padding, and probably the filter window length (the anti-aliasing filter could also be designed to be causal like the one in the acquisition devices, but then the user would have to opt for delaying each signal by some amount).

cbrnr:
I don't understand why you assume that the original signals are not band-limited. Of course they are. So upsampling will not produce anti-aliasing effects. If we downsample, we have to apply an anti-aliasing filter. Or are you saying even in those cases there will be artifacts like the ones from the website (BTW, it'd be helpful if we know what exactly caused these artifacts in each individual example).

chkothe:
So yes, the issue is that in the linear interpolation graphic, there are zig-zags, and those contain sharp edges of much higher frequency than the original signal. The correct interpolation would be a wave that crosses through those points, possibly overshoots in some places, and generally stays below the nyquist rate.

cbrnr:
Can you be more specific? I don't understand what you mean by zig-zags. The examples all perform downsampling from 96kHz to 44.1kHz, and they explicitly state that all problems are due to imperfections of the used anti-aliasing filter.

chkothe:
Ah sorry, so in the upsampling example from earlier, the new signal has corners:
upsample

And in the frequency spectrum, those corners can't be represented by a sum of sinusoids up to the original signal's nyquist frequency. Instead, they have a footprint in the upper part of the spectrum of the new signal, up to that signal's nyquist rate.

chkothe:
... and the same holds even if use quadratic or cubic mode in interp1d. For example, consider this signal of a sine wave (in orange) that was sampled at some rate (the spikes), and we wish to upsample that by 5x (from https://www.wavewalkerdsp.com/2022/09/01/introduction-to-interpolation-and-upsampling/). The figure has zeros filled in already for the samples that we're asking interp to generate for us (4 zeros for every original sample). 

IntroductionToInterpolationAndUpsampling_upsampled

If we were to avoid spectral aliasing, our new sample points have to lie exactly on the orange sine wave. If we deviate, especially if our deviations are a wave "riding on top" of the sine wave (i.e., higher frequency), then we introduced aliasing. So if we put say a smooth sigmoid-like polynomial in between any two successive samples of the original signal (between the non-zero spikes), we will get "wiggles" relative to the sine wave, and so that would be high-frequency aliasing that would manifest in various interesting ways in the spectrum. So we'd need to instead run an anti-aliasing filter over that, same as in the downsampling case, but then we introduce non-causality (and potentially some filter artifacts depending on how that anti-aliasing filter is windowed, how long it is, etc.

cbrnr:
Ah, there's probably a misunderstanding. My little toy example was just meant to illustrate how we could handle gaps in the data. I was using simple interpolation *without* an anti-aliasing filter afterwards. Of course, the actual resampling must include such a low-pass filter, otherwise we'll definitely run into aliasing.

So as I've said previously, proper resampling (meaning that anti-aliasing filters must be included) should not produce the artifacts you posted. In fact, these filters all suffer from very sub-optimal low-pass filters, as you can verify by looking at the frequency responses.

Here is the spectrogram and frequency response for a filter with a very weak low-pass filter (Ableton Live 7):

Ableton-2
Ableton

But here's what a good low-pass filter looks like (Acon Digital Acoustica 7.4):

Acoustica74
Acoustica74-2

This shows that if done correctly (i.e. if the anti-aliasing filter is steep enough), resampling works pretty well.

Of course there are other issues to consider with filtering, since they by definition "distort" the signal. Non-causality should not be a huge problem in most applications, especially if you don't forward/backward filter, but compensate the delay of a FIR filter by shifting the result back.

I agree that resampling is not without problems, but we're not trying to make that the default behavior of `load_xdf()`. We will also include cautionary notes for the new option, alerting people to be careful when they resample. But if you want to process several streams simultaneously, there are not really many alternatives than to resample. This is especially true when there are gaps in the original signal, which led to exactly this discussion here.

Or do you have an alternative suggestion for dealing with gaps in recordings as well as synchronizing different streams so that they can be processed simultaneously?

chkothe:
Got it. Yeah I think with a combination of some cautionary notes and a best-effort approach to resampling (i.e., a decent antialiasing filter, which will have to be fairly long to be steep), and it not being a default, that could pan out. 

Generally, if we can keep users from falling into these traps when they would otherwise roll their own approach, that could help. But it's good to keep in mind along the way that it's really an estimation problem that has several baked-in tradeoffs and parameters inherent to it, some of which a user has to be aware of depending on what analysis they're planning to do (e.g., BCI, connectivity, etc). We could hard-code some of those parameters (e.g., filter length) to keep the API minimal, or we could have a way for users to pass that in when they want to override it (like with a dictionary), although I have a hard time imagining anyone experimenting with different filter parameters in their data loader... It'd seem more likely that they'd do that after having loaded the raw data to explore how that impacts their analysis, basically as part of their preprocessing chain. That's actually one reason why making it a separate function that's part of pyxdf could be a good idea, since it has room for more options and their documentation, one doesn't have to reload the data from disk for each parameter setting one wants to try, and it seems like a more conventional boundary between data loading and preprocessing.

That said, to answer your other q, one of our strategies has been for a long time to only resample & stack multi-stream data when it's necessitated by the analysis, which isn't all that often. For example, oftentimes the streams only end up getting related to each other at the level of derived measures, e.g. from epoched/segmented data (like time/frequency averages), at which point the original samples have already vanished (and with them the need for a sample-by-sample alignment). And even a fair amount of time-domain multimodal analyses like CCA can work with multirate signals, although it's surely easier to work with multiplexed data. In any case though we'd want the authors' paper to say "we resampled the data to a common sampling rate using [insert filter here, e.g., a windowed sinc filter, the `load_xdf` defaults, etc] and then performed XYZ on the resulting stacked time series".

cbrnr:
I'm OK with putting all of this resampling business into a separate function.

Regarding filter settings, let's just use some sane (relatively steep) defaults for now.

And yes, I agree that resampling multiple streams is probably a bit of a niche use case. But originally, we really wanted to tackle the problem of gaps in a stream. It is described here: https://github.com/cbrnr/mnelab/issues/385
How would you deal with this issue? Do you think resampling is appropriate? Maybe we can separate this discussion into two goals: (1) Sync up multiple streams and (2) deal with gaps in a stream.

chkothe:
Yeah a gaps-as-nans option would seem to be useful in `load_xdf`, since it already has logic to detect gaps in the first place, and just doesn't make an effort to organize it all into one gapless time series. If we do assume that the hardware kept sampling on the same clock and we just lost some number of those samples (e.g., packet loss), then we'd just need to insert that number of NaNs into the gap (and the downstream analysis could try to impute that or not, etc). If we assume that the hardware was turned off and back on (and therefore the HW clock was reset), then technically speaking the sample positions after the gap are most likely shifted by a fraction of a sample if we inserted k NaNs. The option could be documented to imply this very consequence - i.e., it would introduce a +/- 0.5 samples uncertainty into the timing of each stream if the stream had gaps _that also_ coincided with a restart of the device's clock. However, it'd have to be implemented in such a way that the errors don't compound across successive gaps -- i.e., it'd have to preallocate the new NaN-filled sample array and then insert the stretches into sample ranges with a rounded start offset.

So now one could try to reduce that temporal uncertainty and do resample that signal onto the new (fractionally shifted) time base  (again, assuming that the device clock really _did_ reset). In a way we're trying to repair data that really wasn't sampled on a regular clock to begin with, and obviously that comes down to a filtering problem again, with all that that implies (kernel lengths, edge effects, pre/post ringing). So if we have a collection of functions for time-domain surgery of data (along the lines of Robert's internal utility functions), then that particular scenario may be better addressed as a special case of the resampling problem, which is more generally also capable of bringing all streams onto the same sampling grid. So there could be the simple mode, which is the correct thing to do and works just fine with things like Bluetooth packet loss or the CPU not keeping up with the webcam etc. (since the device clock never actually resets*), and for everything else the user could be directed to the big and heavy signal surgery function that interpolates all the segments across all the streams, regular or not, whichever sampling rate, onto the same time base (obviously for irregularly sampled streams that's its own little project).

*well sometimes webcams even change the rate mid-way, as some of us found out the hard way.

DominiqueMakowski:
Thanks @chkothe for this very interesting dive-in. 
<sub>*If you have time, would be very happy to have your opinion on the proof-of-concept synchronization implemented [in neurokit](https://github.com/neuropsychology/NeuroKit/blob/3d004b4777c3e59b288f46b9f0f812ac1e04b5d4/neurokit2/data/read_xdf.py#L108-L121) using pandas, from my testing it works pretty well but don't hesitate if you have any comments*</sub>

> That's actually one reason why making it a separate function that's part of pyxdf could be a good idea

+1

> let's just use some sane (relatively steep) defaults for now.

+1, let's pick the worst of all plague to help user get something practically useable as a starting point, and then they can fiddle with the parameters

chkothe:
Sure, thx! I think we could also have some solid unit tests that show us how many dB of error we're adding either in the spectrum if we skimp on the filter, or in time when we shift something that shouldn't be shifted, or due to ringing in the presence of spike artifacts if we have a long filter. That'd help us relate all manner of errors on the same terms (e.g., the infinitewave plots have a pretty unforgiving color scale) and make a good default choice.

agricolab:
Thanks for the feedback. I have three comments.

1. Everyone who recorded a signal close to Nyquist (or has zeros instead of signal intermixed) did something wrong from the start, and should consider scrapping the recording, buy better equipment, and start fresh. IMHO , the typical user wants e.g. EEG sampled at 1kHz in sync with let's say pupil diameter sampled at 100Hz  and a marker stream. Such massively oversampled signals will likely not experience any significant issues with almost all interpolation and resampling methods. Again, if you do, my guess is that you have other problems compared to aligning your samples. 
3. I used linear (and nearest-neighbour when the channel format was string) because they come  builtin or with numpy.
4. Depending on the user needs, they might prefer different interpolation/resampling methods for different reasons. Any default method we could implement would not be adequate to a tricky DSP issue in the recording, so imho there's no need to even try. Consider that some channel formats are int or str, and in those cases interpolation/resampling is even trickier. Instead of creating a automagic and powerful DSP toolbox out of pyxdf, a simpler solution would be to allow the user to specify the interpolation/resampling function for every stream (as string for a scipy or numpy  method, or custom-written callable), maybe as a dict or other map type.

cbrnr:
I agree will all points @agricolab! But we need a sane default for the resampling function. What would you choose?

agricolab:
Hm. My initial idea was to keep it simple, stay ignorant to possible causes of gaps and jitter and offer what's built-in anyways or simple to include with scipy. 

I can understand that we would want to offer one or two more dedicated routines to address some of the issues mentioned by @cbrnr e.g. gaps due to machine resets or non-stationary sampling rate (webcam)

I can also agree that we could move this utility out of load_xdf. We didn't do it with other utility functions, though, but those were apparently simpler. So we could break this big thing into smaller utility functions. Or externalize it to highlight its complexity and non-defaultness.

Two options: a new class XDFlist is returned as first element of the tuple that has the method `align` or similar, or a free function that takes the classical list of defaultdixt as argument. I'd prefer a free function for reasons as discussed earlier.

There is one catch. We have to propagate information about start and stop of a segment outside of load_xdf. Filling gaps with NaNs is probably not recommendable as discussed earlier, i.e. dealing with unequal clock offsets between segments means we would have to smoothen timestamps and either resample or introduce non-stationary sampling rates, but we don't want to do that within load_xdf. Filling with NaNs can also introduce large memory costs.

My preferred option would be to add a key with that information to the stream dictionary, e.g.`gaps`: a list of int tuples (start,stop).

agricolab:
I think there are several things at the same time we want to address.

1. Fix clock resets 
2. Fix non-stationary sampling rates between segments
3. Align samples of multiple streams with different channel formats
4. Resample for other reasons, e.g. memory.

chkothe:
Hey, great comments -- and yes, apologies for bombarding you with all that cascade of requirements! And thanks for stepping up to do something (even if simple) about it. I also agree that a free function would seem cleanest and introduces less coupling and fewer spurious types like XDFlist etc. I also think that it'd be misplaced to try to turn pyxdf into some sort of bottom-up multimodal preprocessing library (it'd be a big feature creep magnet that will turn pyxdf into an unholy monster), so we'd want to know where to draw the line. 

On that, I like your proposal of a pluggable callback function, that sounds like a sensible direction; taking a step back, one could say that there are some tasks that are clearly better handled by the user's favorite preprocessing package (how exactly to resample or interpolate a signal seems to match that, and when one considers that different streams would call for different treatment, e.g., string or integer markers, irregular but smooth, and so forth, that's even more obvious). On the other hand, what seems like squarely in the department of pyxdf is anything that's entangled with determining what's a stretch of signal and what's a gap, what the best estimate of the device sampling rate during a given stretch and where exactly does the stretch begin/end, and might we be dealing with a device clock reset across a gap or not -- just because that happens to be XDF's unique territory, since it's a record of what happened during data collection and not an idealized abstraction thereof (different starts/ends per modality, maybe gaps, sampling rate imperfections).

To me the NaN business does seem like a decent fit for pyxdf, basically in the sense that pyxdf manages the conversion from the "real world" captured in the file to the idealized multi-channel matrix that the user's favorite processing tool expects, but no, we should probably not attempt to solve everything at once. That said, it does become entangled when one wants to resample multiple streams to the exact same time points (where some streams don't have data). Here is something that one could perhaps do (rough off-the-cuff sketch):
- both the resampling problem and the NaN generation problem have in common that we have 1) a series of M time points at which we want to re-estimate the value of a signal,  2) the signal to resample along with its current time stamps (which are piecewise regular after pyxdf has done its first pass of dejittering segments) and perhaps 3) the list of segment start/end times for the segments (the gaps would be implied by that so we wouldn't really need those too)
- we could have function that the user can override which takes 1, 2 and optionally 3 and which does the resampling business. If the user is comfortable with the caveats, this could be numpy.interp1d (takes 1 and 2), or a wrapper for that that first generates M nans and then does numpy.interp only within the segments (i.e., the gaps stay nan's). 

There are 3 ways the time points could be chosen:

a) the time points could be a regular grid at the nominal sampling rate of the highest-rate stream, stretching from a start to end time stamp in the file (the end result would be an idealized multi-channel matrix across all modalities, where everything got resampled, including the highest-rate stream, and there can well be a series of nan stretches if there were gaps, or where streams didn't start a the same time; the user may then trim the start/end to the part that's common to all streams).
b) the time points could also be any desired lower sampling rate, e.g., if there's a very high rate stream in the file like audio.
c)  it could be the time stamps of a chosen "primary" stream (e.g., neural data), in which case we don't resample the primary stream at all to keep it pristine.

The user could also pass in (or we could pre-define, perhaps in a follow-up PR) a wrapper that invokes not numpy but scipy.resample for each segment (as per the optional third callback function argument), and which balks if the target time stamps are not perfectly regular since resample can't handle that case, so this method would be usable in all cases a), b) and c) BUT not c) when the primary stream also had gaps (well if we wanted to cover all conceivable cases, one could partition the operation to run on every regularly sampled stretch of the desired (target) time stamps, assuming it's piecewise regular, but that's definitely material for a future PR). Since resample wouldnt handle fractional sample shift, it would have to either sweep that under the rug (maybe acceptable if primary stream is 500+ Hz but not below), or it might resample to an even higher rate and then subsample at the desired time points (without running out of memory, e.g., done channel-by-channel). I think that instead of calling scipy.resample() on stretches of piecewise regular data someone could probably write a resample function that works on irregular source and target time points but follows DSP best practices (it may be along the lines of using a cheaper interpolator and then low-pass filtering the result).

So the API could then be what you want to resample the data to -- a, b, c, or no resampling? And then an optional callback function or a dictionary of functions keyed by the stream (e.g., type/modality) to do the actual work. There could be one or more predefined functions for that -- pure numpy (at own risk but simple and handles all cases), scipy (if you have it, then using resample), and maybe someone has time to make a PR that provides as a third option an irregular-rate resampler with antialiasing filter (bonus points for only requiring numpy).

Take this sketch with a grain of salt since it's pretty late here. I'm also by no means committed to that, so feel free to take it apart (it's just the first way that came to mind).  A question to ask though is whether there's anything of substance left to do for a high-level framework function that accepts the callback, since user could just as well simply call the callback themselves with either some np.linspace timestamps of their choosing, or with their primary stream's time stamps as returned by load_xdf. But it would mean that the user has to follow a (short) recipe, maybe in the form of some example script, and isn't simply calling a one-liner function with some arguments.

agricolab:
Thanks @chkothe for the extensive feedback

I like the idea to select a primary stream as alternative to setting the sampling rate or picking the fastest stream. The challenge with using the timestamps of a primary stream and not touching is timeseries was that this only works if there are no segments/gaps. 

Everything complex that works grew from something simple that worked. 

So my suggestion is that we offer to fix gaps by filling with NaNs at irregular timestamps that smoothen the offset between segments.

Then we offer the option to align timestamps if specific conditions are met.

An excourse: my practical experience is that most users don't care much about timestamps besides matching them with markers. If the EEG amplifier says it samples with 1kHz, but the xdf says the actual effective srate is 1001.0576 most users would just forward the array to mne or whatever  as if it were sampled at 1kHz.

That creates the idea that we might consider to use shift-alignment using nearest-neighbour, as it comes without resampling and any risk of distortion. 
It means that we get a list of new regularized timestamps and create the timeseries by picking the closest sample from the old timestamps.

Only if that fails because a original timestamp is the closest neighbour of two or more new "regularized" timestamps, interpolation or resampling would be necessary.

This approach has imho the advantage that it formalizes the assumptions behind what users usually do anyways, and fails, if these assumptions don't hold. The approach would succeed more likely when the xdf files are small, and where streams have no gaps or gaps could be fixed, and the sampling rate of all streams to be aligned is approximately the same.

I'm not sure what to introduce the option to resample. Most likely the best solution is that it should be done by the user who can then try the resample individual streams and then calls `align_streams` function again. But on the other hand, alignment can be forced with resampling.

So how about two free functions. One that only works under specific conditions and uses nearest neighbour. That one could even live within load_xdf.

 And a second one that aligns for sure using  interpolation/resampling (as defined stream wise and using a list of common timestamps as given by the user) but might introduce distortion? And yes that's just a wrapper function, but might perform some sanity checks.

agricolab:
Thanks @chkothe for the extensive feedback

I like the idea to select a primary stream as alternative to setting the sampling rate or picking the fastest stream. The challenge with using the timestamps of a primary stream and not touching is timeseries was that this only works if there are no segments/gaps. 

Everything complex that works grew from something simple that worked. 

So my suggestion is that we offer to fix gaps by filling with NaNs at irregular timestamps that smoothen the offset between segments.

Then we offer the option to align timestamps if specific conditions are met.

An excourse: my practical experience is that most users don't care much about timestamps besides matching them with markers. If the EEG amplifier says it samples with 1kHz, but the xdf says the actual effective srate is 1001.0576 most users would just forward the array to mne or whatever  as if it were sampled at 1kHz.

That creates the idea that we might consider to use shift-alignment using nearest-neighbour, as it comes without resampling and any risk of distortion. 
It means that we get a list of new regularized timestamps and create the timeseries by picking the closest sample from the old timestamps.

Only if that fails because a original timestamp is the closest neighbour of two or more new "regularized" timestamps, interpolation or resampling would be necessary.

This approach has imho the advantage that it formalizes the assumptions behind what users usually do anyways, and fails, if these assumptions don't hold. The approach would succeed more likely when the xdf files are small, and where streams have no gaps or gaps could be fixed, and the sampling rate of all streams to be aligned is approximately the same.

I'm not sure what to introduce the option to resample. Most likely the best solution is that it should be done by the user who can then try the resample individual streams and then calls `align_streams` function again. But on the other hand, alignment can be forced with resampling.

So how about two free functions. One that only works under specific conditions and uses nearest neighbour. That one could even live within load_xdf.

 And a second one that aligns for sure using  interpolation/resampling (as defined stream wise and using a list of common timestamps as given by the user) but might introduce distortion? And yes that's just a wrapper function, but might perform some sanity checks.

cbrnr:
From an MNE perspective, I need a regularly sampled 2D NumPy array with channels in rows and samples in columns. Originally, I was planning to implement the conversion from loosely synced XDF streams (as returned by `load_xdf()`) outside of pyxdf. But then I thought that the majority of users will probably need to do that as well, so it would be a shame if every package had to implement that.

I'm all for avoiding feature creep though, but some way to get a 2D array of all streams, synced in time, where gaps have been filled with NaNs, would be nice. If this is doable with pluggable callables, great! However, if we use `np.interp()`, we immediately run into the problem that @chkothe pointed out previously. We definitely need to include an anti-aliasing filter.

Can we try to come up with a plan for a minimal implementation? I mean, options range from (1) doing nothing and leaving everything to the user to (2) implementing a whole bunch of functions that take into account all the edge cases we can think about. I think we agree that we want neiter of these extreme options, but what exactly can we agree upon?

cbrnr:
(@agricolab I deleted your comment, because it was a duplicate of your last comment.)

agricolab:
I can see that the moment we integrate a method for interpolation/resampling  as default,, we run into a risk for the users and the developers. Users might naively stick to the default whether it is sensible or not, and the developers might be faced with accusations and feature creep.

The challenge is now that when we want to create an aligned 2D  array from all selected streams where the timestamps are not already close to each other (i.e. can be aligned by shifting), we need to interpolate or resample.

One option is that we do not support a default method for resampling, but require that the user **ALWAYS** supports one (e.g. a callable). We would not implement this as kwarg to load_xdf to prevent any confusion on user responsibility and allow to explore different resampling methods without heavy file access cost.

All we need to do for that would be 
a) add a field with start and stop of gaps/segments to the defaultdict stream object returned by load_xdf
b) implement a free function to align streams using that resampling method

That free function would align samples and fill gaps with NaNs. If no interpolation method is supported by the users, the function will attempt to align all streams by shifiting and fail with a warning or exception if it can't. All streams for whom  an interpolation method are supported will be aligned using that method, and the function will fail if the uninterpolated streams still can't be aligned.

We write extensive documentation, e.g. on how to create callables for the align-function, e.g. with lambdas or `__call__`, some examples with more or less sensible options for some typical channel formats (e.g. EEG), and warn of the risks involved in resampling. 

We could also implement a factory function, that returns a callable given a few kwargs (using scipy or mne or whatever as backend). These might actually  be better situated in a different project, e.g. a to be created `xdf-aligners` or mne or somewhere else.

cbrnr:
I like it! As long as we have some nice examples on how to do it with typical signals, this is a perfect solution!

agricolab:
Would @chkothe  also be comfortable with this solution?

agricolab:
Great, i'll start working on a free function as discussed. If somebody else wants to update the documentation, that'd be great. 

I assume that every callable would take three arguments `foo(timestamps, timeseries, new_timestamps)`, as the current `_interpolate` https://github.com/agricolab/xdf-Python/blob/sync_timestamps/pyxdf/pyxdf.py#L911

chkothe:
Yeah -- I think it could also have an optional 4th argument like `segments` (list of start/end indices of fixed-rate segments among the timestamps) so that that callable doesn't have to try to detect gaps in `timestamps` itself if it wants to use a method that requires fixed-rate inputs. Btw what do you think about `old_timestamps` and `old_timeseries` to differentiate that more clearly (esp once we add something like segments, which might as well go with the new timestamps if those were chosen to be the ones from the fastest-rate stream, which may have gaps too)? Might feel like an abundance of arguments in the user's callable, but then they can always choose to ignore them if they don't need them...

agricolab:
I agree, `old_timestamps` and `old_timeseries` is certainly better.

My suggestion for the function hierarchy would be similar to:

Segment start and stops (as detected by _jitter_removal) within load_xdf will not be discarded anymore, but included in the defaultdict stream objects in their own field (let's say `stream["segments"]`) as a list of tuples of ints (that is pairs of sample indices).

The free function, lets say `align_streams(streams, align_foo:Dict = None,  use_time_stamps:List[float] = None)`, which takes the streams as returned by `load_xdf` and (optionally) a dictionary of Callables with streamid (int) as key, and (optionally) timestamps to be used for alignment/interpolation/resampling.

The function will initialize a 2Darray of the expected size and fill it with NaNs. For each stream, this function will then  use the callable each  segment-wise, with segments as defined in the stream object.

chkothe:
That all sounds about right (returning segments, defaulting to eff. srate of fastest stream [_if the stream has no gaps btw then presumably it doesn't have to be resampled, and if it does have gaps, then maybe segments in it still only need to be shifted -- there could be an option whether the user wants that or resample it_], using the total time range of the data, etc). Applying the callable per segment in `use_time_stamps` seems reasonable too (note though you'd have to detect those gaps in there if someone passes `use_time_stamps=streams['eeg'].timestamps`, but arguably that's much easier since those should generally be segment-wise dejittered stamps). Or you might have the option to specify a string key in streams, like use_time_stamps='eeg', because then you could just look up the segments metadata in streams['eeg')? Then you can save yourself that detection and you could require that use_time_stamps, if user passes an array, can only be perfectly regular.

For string marker streams, assuming those don't enter the 2D array by default (one could write a row with indices into a string table but seems like a lot of work, more API surface, and it may not really be the format that the user's software wants for markers anyway).

chkothe:
Thanks for pushing it forward! Yes, seems best to not rush it now since it'll be hard to change the API later (and much of it seems to be API surface after all).

agricolab:
Ok. I implemented shift_alignment as a default, and handle segments. Unit and system tests run fine for segmented and unsegmented streams, and when `aligned_timestamps`, `sampling_rate` or interpolation function is supported by the user.  See e.g. system tests with minimal.xdf here: https://github.com/agricolab/xdf-Python/blob/sync_timestamps/pyxdf/test/test_data.py#L80-L123)

So i guess it's time to let you fellows play with it. Looking forward to your feedback. Hope i didn't miss a big thing...

agricolab:
In regards to 
>note though you'd have to detect those gaps in there if someone passes use_time_stamps=streams['eeg'].timestamps
I made my life easy. If a user support timestamps that do not cover a span where a stream has samples, shift_alignment will fail with an exception. Therefore, this approach would only work if an interpolation function is supported that can handle this. In that case, it doesn't really matter whether the `new_timestamps` have gaps or not.

chkothe:
Ok great -- thanks for pushing through! Will have a look this weekend at the latest.

cboulay:
`load_xdf` has docstrings for the new arguments but these aren't yet used by `load_xdf`. Is that intentional?
So for now you're just looking for feedback on `align_streams`?

(Incidentally, is it necessary that `use_samplingrate` is int or str? What about float or str?)

agricolab:
> load_xdf has docstrings for the new arguments but these aren't yet used by load_xdf. Is that intentional?

sorry, that was leftover from an earlier implementation. `load_xdf` should be untouched.

> So for now you're just looking for feedback on align_streams?

yes, and it's helper functions.

agricolab:
I refactored the main alignment function `align_streams` and the helper functions into their own file for better maintainability, and added basic docstring.  `align_streams` will be imported to pyxdf.

cbrnr:
@agricolab I tried to test this new functionality, but I'm not sure how exactly I'm supposed to use it.

For example, I have a data set with a large gap somewhere in the middle of the recording (available [here](https://osf.io/download/6y8ub/)). I would like to load this file and then sync and resample all streams to 256Hz. Then I would like to end up with a 2D NumPy array containing all EEG channels. Is this possible? If so, how? The problem here is that with my resampling implementation in MNELAB, I'm completely ignoring any gaps in the data (https://github.com/cbrnr/mnelab/issues/385).

agricolab:
Talk about revenants... Oh boy, have to take a look myself after a year or so...

So, we have the free function `align_streams` see https://github.com/agricolab/xdf-Python/blob/sync_timestamps/pyxdf/align.py#L91C5-L91C18 that takes 
1. streams as returned by load_xdf. 
2. a dictionary that maps streamids to respective interpolation functions (default is `_shift_align`)
3. user-defined aligned_timestamps (defaults to earliest of all streams to latest of all streams)
4. user defined sampling rate (defaults to sampling rate of the fastest stream

`_shift_align` simply moves the sample to the closest matching new timestamp. IIRC we discussed this extensively, and there is no default resampling involved. If samples can not be assigned (because of effective downsampling or edge cases), the default approach is to fail with a run time error.

If you need another interpolation function that can handle resampling, you have to specify the second argument, and assign a callable to those streamIds that need resampling. The module supports in addition to the default `_shift_align` also `_interpolate` which is a wrapper for scipy.interpolate unless this dependency is not met, than it wraps numpy.interpolate.

So at least from reading code and docs and memory, you would need to support:

The streams, the sampling rate (256 in your case), and a callable that can handle downsampling.

That leaves open your question on **how are gaps handled**. 

First, gaps are detected as defined by load_xdf IIRC i had to change load_xdf to export this, so it might now work with recent load_xdfs. Second, each aligned timeseries is initialized with NaN and only segments with actual samples are aligned/interpolated, leaving gaps to stay NaNs. See https://github.com/agricolab/xdf-Python/blob/sync_timestamps/pyxdf/align.py#L162-L175

cbrnr:
Thanks @agricolab, unfortunately it doesn't work for me using [this dataset](https://osf.io/download/6y8ub/) and the following code:

```python
import pyxdf

fname = "sub-13_ses-S001_task-HCT_run-001_eeg.xdf"
streams, header = pyxdf.load_xdf(fname, select_streams=[2])  # EEG stream

pyxdf.align_streams(streams)
```

This results in the following error:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/clemens/Projects/pyxdf/pyxdf/align.py", line 165, in align_streams
    _new_timeseries = align(
                      ^^^^^^
  File "/Users/clemens/Projects/pyxdf/pyxdf/align.py", line 75, in _shift_align
    raise RuntimeError(f"Too few new timestamps. {missed} of {len(old_timestamps)} old samples could not be assigned.")
RuntimeError: Too few new timestamps. 1 of 43848 old samples could not be assigned.
```

Not sure if this is supposed to work with one single stream in the first place (which would be extremely important for me), but I also tried with two streams which also yield an error:

```python
import pyxdf

fname = "sub-13_ses-S001_task-HCT_run-001_eeg.xdf"
streams, header = pyxdf.load_xdf(fname, select_streams=[2, 5])  # EEG and ACC streams

pyxdf.align_streams(streams)
```

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/clemens/Projects/pyxdf/pyxdf/align.py", line 165, in align_streams
    _new_timeseries = align(
                      ^^^^^^
  File "/Users/clemens/Projects/pyxdf/pyxdf/align.py", line 81, in _shift_align
    toomany[old_timestamps[source[v]]].append(new_timestamps[target[n]])
                                                             ~~~~~~^^^
IndexError: list index out of range
```

agricolab:
Hm. Sorry to hear. I'll take a look asap, probably this weekend.

agricolab:
Okay, no longer Python 3.12 compatible, because pkg_resources is deprecated... Guess that's another issue. Fixed it by wrapping it into a try-catch.

# First issue

Your first issue stems from the fact that there is no unique mapping from the old to the new timestamps. Likely due to down-sampling.  The bug is in the reporting, where i check how many new timestamps are possible for each old timestamp. This is not really relevant, i fixed it for the time being with a catch IndexError.

The gist is https://github.com/agricolab/xdf-Python/blob/8fa4645f50904dff6f3fe00ada57c06d1c87d6dc/pyxdf/align.py#L76 `if len(set(source)) != len(source):  # non-unique mapping, will throw` 

In consequence, because your two streams can not be aligned by shifting (because multiple new timestamps are the closest neighbour of the old timestamp), you would need to resample to align both streams.

cbrnr:
> Okay, no longer Python 3.12 compatible, because pkg_resources is deprecated... Guess that's another issue. Fixed it by wrapping it into a try-catch.

If you rebase onto `main`, that problem has already been fixed.

agricolab:
Yeah, but if i rebase, i'll have to fix merge conflicts... :D

agricolab:
Okay, at first i thought that the second issue stems from some edge case, i..e that new_timestamps older or younger than any old timestamps (i.e. the original timestamps) could not be assigned. That didn't seem to be the problem. 

Instead it appears that if you do not support timestamps or sampling rate yourself, the function picks it from the stream and creates a new, aligned_timestamps. Now, we face some floating errors


```
    if aligned_timestamps is None:
        # we pick the oldest and youngest timestamp of all streams
        stamps = [stream["time_stamps"] for stream in streams]
        ts_first = min((min(s) for s in stamps))
        ts_last = max((max(s) for s in stamps))
        full_dur = ts_last - ts_first
        step = 1 / sampling_rate
        # we create new regularized timestamps
        aligned_timestamps = np.arange(ts_first, ts_last + step / 2, step)
        # using np.linspace only differs in step if n_samples is different (as n_samples must be an integer number (see implementation below).
        # therefore we stick with np.arange (in spite of possible floating point error accumulation, but to make sure that ts_last is included, we add a half-step. This therefore comes at the cost of a overshoot, but i consider this acceptable considering this stamp would only be from one stream, and not part of all other and therefore is kind of arbitray anyways.
        # linspace implementation:
        # n_samples = int(np.round((full_dur * sampling_rate),0))+1
        # aligned_timestamps = np.linspace(ts_first, ts_last, n_samples)```
        
        i even had that commented apparantly

**Comment by agricolab** [comment]
So, maybe it comes from the regularization, i.e. the original timestamps miss the beat somehow, and have an additional sample at a weird timestamp strewn in; or it comes from an interaction between the new aligned_timestamps and having segments. 

Maybe i find the time in the weekend...

**Comment by agricolab** [comment]
Okay, i inspected the data a little bit more and it seems _shift_align can not assign an old_timestamp from the middle of the segment
```
RuntimeError: Too few new timestamps. 1 old timestamps ([182337.61104443]) found no corresponding new timestamp because it was already taken by another old timestamp.
```

shift_align goes through all new_timestamps and finds the closest old_timestamp (unless it is already mapped to another new_timestamp. That appears to be the case here. As the new_timestamps (whether created by linspace or arange) throw in both cases (but can be assumed to be unique and more or less correctly spaced, i would assume the issue stems from the fact that timestamps were created differently.

By default., pyxdf dejitters timestamps, and this happens with these streams too (othwise we wouldn't have segments). But jitter_removal factually uses different sampling rate for different streams, i.e. it linearizes from first to last timestamp (using linalg.lstsq). But the effective_srate is calculated across all segments.

align_streams instead creates new_timestamps evenly spaced across the whole span from the very first to the very last timestamp of the stream.

So, i would assume the issue lies there. A small misalignment makes it impossible to shift_align within a segment, because effective srates do not align sufficiently well. 

In general, i should note that shift_align is expected to fail in most cases, because it has very strict requirements...

**Comment by agricolab** [comment]
When i interpolate instead (i.e.

```
streams, header = pyxdf.load_xdf(fname, select_streams=[2])  # EEG stream
align = {2: lambda x, y, xh: pyxdf.align._interpolate(x, y, xh, "cubic")}
aligned = pyxdf.align_streams(streams, align_foo=align)
```

it passes, but alas, performs interpolation...

**Comment by cbrnr** [comment]
Thanks @agricolab, now this makes sense! With thorough documentation (including examples), I think this will be a nice addition!

However, I still have questions 😄:

Maybe my understanding of pyxdf is too limited, but what exactly are those segments you mention? Are they available in the output? And if the effective sampling frequency ignores segments, it is not really reliable in this case, right?

I'm viewing this problem from a user perspective. Basically, I have some streams, and pyxdf gives me those streams on a common time basis. However, I also need them on a uniform time grid, which means that these streams need to be resampled, even if it's just a single stream, but most definitely for two or more streams. For a single stream, I would like to be able to take the samples without any further resampling and just assume the nominal (or effective) sampling frequency is accurate. This is now possible with your `pyxdf.align_streams()` function, right? (The parameter name should probably be improved.)

**Comment by agricolab** [comment]
> I would like to be able to take the samples without any further resampling and just assume the nominal (or effective) sampling frequency is accurate.

Easiest is to not use any alignment then at all. Simply discard the information contained int the timestamps, and stick with your assumption. Did it all the time with my pure EEG streams, and it works usually sufficiently reliable if the duration of the recording is not too long. Especially if you use medical grade EEG amplifiers, one can usually assume that no samples are dropped, and sampling rate is sufficiently constant, and assume that all jitter stems from TCP/IP or other transporting or protocol translations.

Now, more generally, we face two problems. First, segments. Second, aligning multiple streams with different clocks (typically markers and EEG).

Segment can occur when there is a break due to hardware or network issues, and data is delayed or missing. It then depends on your additional knowledge or assumptions. If samples are missing, we are really talking about a gap between segments. If it is just delayed in transport, you might want to ignore the gap and assume constant sampling rate. If you had to change amplifiers during the recording, you might actually have a slightly different clock now. LSL/XDF couldn't really know. All it offers are methods to dejitter and segment streams if gaps are too large. This information about segments is relayed to align_streams.

Second, because each stream has its own clock and jitter, no two samples from different streams really perfectly align. Markers are typical examples, and we might want to know the closest sample in another stream. Or we might want to align different physiological signals. LSL/XDF couldn't really know. So you can decide whether you only want to shift the data (works usually quite well with a Marker stream) or actually interpolate. The latter means you have to rely very much on the information in the timestamps (or manipulate it before you pass it to align_streams)

Now, back to the start. If you have a single, non-segmented stream, and believe that the sampling rate should be uniform, just discard the timestamps and take the manufacturers sampling rate as gospel.

**Comment by cbrnr** [comment]
Thanks @agricolab, that's very reasonable! Just one more minor question, how do I know that pyxdf detected segments? Is this reflected in the output somehow?

**Comment by agricolab** [comment]
Jupp, you can control detection with jitter_break_threshold_* (i.e. https://github.com/xdf-modules/pyxdf/blob/main/pyxdf/pyxdf.py#L165C1-L173C74) 

IIRC for some reason, pyxdf exposes this not by default. I had to change https://github.com/agricolab/xdf-Python/blob/sync_timestamps/pyxdf/pyxdf.py#L405C9-L405C36 to actually pass this info. IMHO, should be default :D

**Comment by agricolab** [comment]
Oh, and obviously, you vcan turn dejittering and therefore segmenting completely off  by setting `dejitter_timestamps` to False https://github.com/xdf-modules/pyxdf/blob/main/pyxdf/pyxdf.py#L122

**Comment by cbrnr** [comment]
Interesting, I've never paid attention to those arguments. But how does the output structure differ if there are segments? The docs don't mention that...

**Comment by agricolab** [comment]
That's the point. AFAIK it doesn't. Only the fork from this PR exposes it explicitly in the info field so that align_streams can use that.

**Comment by cbrnr** [comment]
:surprise:

**Comment by behinger** [comment]
I ran into issues with the xdf-interpolation in mnelab and I prefer the apporach by @agricolab to not interpolate if not necessary. What is the status of this PR?

(no hurry: For now I will simply copy the `align.py` function + some custom code to create the raw files)

**Comment by agricolab** [comment]
Thanks for the feedback. 

For XDF files that large, a unique shift alignment is unlikely to work out anyways :D

Joking beside, `_shift_align` was not really written with performance in mind... 

The functions create a lot of lists (e.g. source,  target, too_old, too_young) and makes copies of the timeseries (constructing an array from the timeseries)

These should likely be `old_timestamps = np.asarray(old_timestamps)`  instead of `old_timestamps = np.array(old_timestamps)` etc.

The function makes a full distance vector for each sample (see `  closest = np.abs(old_timestamps - nts).argmin()`: The distance vector is only temporary, but  I am not sure how the python gc manages such temporary objects.

One could also switch to using generators instead of lists (i.e. `(i for i in range(len(old_timestamps)) if i not in source)`  instead of `[i for i in range(len(old_timestamps)) if i not in source]` (just be wary to restart the generators, e.g. for 
```
stamps = [stream["time_stamps"] for stream in streams]
ts_first = min((min(s) for s in stamps))
ts_last = max((max(s) for s in stamps))
```
iirc a generator would be empty when called again for ts_last


And finally, calls to np.unique or argmin can also make copies if called on a list...

So, i would also start adding typehints and mypy to see where type conversions occur....

Or - we could switch to a C/C++ extension



In either case, i likely do not have the time to work on optimization in the next weeks... but i can review or support

behinger:
Thanks for the insights. Unfortunately, I already spend too much time getting something simple to work. I even considered switching to matlab (!). While I think the alignment is the correct solution for most of the data I'll see, I now use the mnelab interpolation, from ~1000Hz to 1000Hz, with subsequent downsampling to 500Hz. Turns out, that merely exporting a `mne.raw` requires you to hold the same dataset at least twice in memory.

I havent looked into any memory details, but I'm still unsure how a 3GB xdf can result in >35GB just to import, interpolate and export again :/. The xdf is float32, then MNE -> float64, ok at 7GB. One would expect that it's easy to save a 7GB EEG dataset with 40GB ram.. but no, not as BrainVision, or as EDF. I'm saving now as eeglab, it seems the most memory efficient implementation. Crazy.

Unfortunately, MNE doesnt support float32 as far as I could find out, so downsampling it is... I'll probably rather improve my LSLTools.jl (based on your LSL.jl) to support segments, rather than try to fiddle with python / C/C++. I also dont know enough python to really deal with memory overhead.

---

Sorry rant over. :(

cbrnr:
I think optimizations are important for this addition, especially since the current implementation does not seem to be feasible for some (many?) use cases.

@behinger could it be that you're mixing up several things in your example? How is exporting to BrainVision or EDF related to this PR (or pyxdf in general)? I understand that you are frustrated by the current situation, but such comments are really not helpful for us developers, as they contain very little information about what the actual problem is.

behinger:
(You are right, I removed it, the relevancy of course exists: if anyone wants to use pyxdf to import data and export it again in a different format, but needs 4x as much memory to do so, then this is a problem - but it is not directly the problem of this PR, you are right)

cbrnr:
Can you report the memory export issue (preferably with a minimal reproducible example) in the relevant repository (presumably MNELAB, but if it's MNE-Python I can forward it)?

PR #2, opened 2019-04-13
Select formats based on index

tstenner:
As a precursor for the chunk7 reader, use the same format ordering as in lsl and read in the native datatype (line 293)

### DIFF
--- a/pyxdf/pyxdf.py
+++ b/pyxdf/pyxdf.py
@@ -5,6 +5,7 @@ This function is closely following the load_xdf reference implementation.
 Copyright (c) 2015-2018, Syntrogi Inc. dba Intheon
 """
 
+import collections
 import os
 import struct
 import itertools
@@ -170,10 +171,15 @@ def load_xdf(filename,
         """Temporary per-stream data."""
         def __init__(self, xml):
             """Init a new StreamData object from a stream header."""
-            fmt2char = {'int8': 'b', 'int16': 'h', 'int32': 'i', 'int64': 'q',
-                        'float32': 'f', 'double64': 'd'}
-            fmt2nbytes = {'int8': 1, 'int16': 2, 'int32': 4, 'int64': 8,
-                          'float32': 4, 'double64': 8}
+            fmts = collections.OrderedDict([
+                ('double64', np.float64),
+                ('float32', np.float32),
+                ('string', np.object),
+                ('int32', np.int32),
+                ('int16', np.int16),
+                ('int8', np.int8),
+                ('int64', np.int64)
+            ])
             # number of channels
             self.nchns = int(xml['info']['channel_count'][0])
             # nominal sampling rate in Hz
@@ -194,10 +200,11 @@ def load_xdf(filename,
             self.tdiff = 1.0 / self.srate if self.srate > 0 else 0.0
             # pre-calc some parsing parameters for efficiency
             if self.fmt != 'string':
+                # index of the format as in channel_format_t
+                self.fmtindex = list(fmts.keys()).index(self.fmt) + 1
+                self.dtype = np.dtype(fmts[self.fmt])
                 # number of bytes to read from stream to handle one sample
-                self.samplebytes = self.nchns * fmt2nbytes[self.fmt]
-                # format string to pass to struct.unpack() to handle one sample
-                self.structfmt = '<%s%s' % (self.nchns, fmt2char[self.fmt])
+                self.samplebytes = self.nchns * self.dtype.itemsize
 
     logger.info('Importing XDF file %s...' % filename)
     if not os.path.exists(filename):
@@ -288,7 +295,7 @@ def load_xdf(filename,
                                 values[k][ch] = raw.decode(errors='replace')
                     else:
                         # read a sample comprised of numeric values
-                        values = np.zeros((nsamples, temp[StreamId].nchns))
+                        values = np.zeros((nsamples, temp[StreamId].nchns), dtype=temp[StreamId].dtype)
                         # for each sample...
                         for k in range(nsamples):
                             # read or deduce time stamp
@@ -300,7 +307,11 @@ def load_xdf(filename,
                             temp[StreamId].last_timestamp = stamps[k]
                             # read the values
                             raw = f.read(temp[StreamId].samplebytes)
-                            values[k, :] = struct.unpack(temp[StreamId].structfmt, raw)
+                            # no fromfile(), see
+                            # https://github.com/numpy/numpy/issues/13319
+                            values[k, :] = np.frombuffer(raw,
+                                                         dtype=temp[StreamId].dtype,
+                                                         count=temp[StreamId].nchns)
                     logger.debug('  reading [%s,%s]' % (temp[StreamId].nchns,
                                                             nsamples))
                     # optionally send through the on_chunk function
--- a/pyxdf/pyxdf.py
+++ b/pyxdf/pyxdf.py
@@ -5,7 +5,6 @@ This function is closely following the load_xdf reference implementation.
 Copyright (c) 2015-2018, Syntrogi Inc. dba Intheon
 """
 
-import collections
 import os
 import struct
 import itertools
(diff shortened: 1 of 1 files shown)


cbrnr:
Does this also do what https://github.com/sccn/xdf/pull/38 proposes? If not, it would be nice if you could integrate that into this PR.

tstenner:
> Does this also do what sccn/xdf#38 proposes?

Yes, I didn't see that but it does everything sccn/xdf#38 does (and a tiny bit more). Plus, it's already in the correct repository :-)

cbrnr:
OK great, that's what I wanted to know - so we can safely close https://github.com/sccn/xdf/pull/38 once this gets merged.

pyxdf/pyxdf.py
```diff
@@ -170,10 +170,10 @@ class StreamData:
         """Temporary per-stream data."""
         def __init__(self, xml):
             """Init a new StreamData object from a stream header."""
-            fmt2char = {'int8': 'b', 'int16': 'h', 'int32': 'i', 'int64': 'q',
-                        'float32': 'f', 'double64': 'd'}
-            fmt2nbytes = {'int8': 1, 'int16': 2, 'int32': 4, 'int64': 8,
-                          'float32': 4, 'double64': 8}
+            fmts = ['double64', 'float32', None, 'int32', 'int16', 'int8', 'int64']
```
cboulay: A dict seems like a more natural container here.

`fmts = {'double64': {'char': 'd', 'nbytes': 8, 'dtype': np.float64}, 'float32': 'char': 'f', 'nbytes': 4, 'dtype': np.float32}}`

With better white spacing, the above is easier to understand and use in the code below. It's easier to maintain in that you don't have to pay attention to ordering in 4 different lists, and, while typing keys is admittedly a downside, doing it incorrectly should generate an error immediately, whereas incorrect ordering will be a silent bug.

cbrnr: Agreed!

tstenner: Ok, guess I'm overruled :-)

In order to make things simple and foolproof, I have replaced out the call to `struct.unpack` with `numpy.frombuffer` so we don't need the char argument. I'd have taken `numpy.dtype.char`, but struct and numpy have different opinions what character int64 should have.

fmts is now an OrderedDict so we get the benefits of both worlds.

pyxdf/pyxdf.py
```diff
@@ -300,7 +307,11 @@ def __init__(self, xml):
                             temp[StreamId].last_timestamp = stamps[k]
                             # read the values
                             raw = f.read(temp[StreamId].samplebytes)
-                            values[k, :] = struct.unpack(temp[StreamId].structfmt, raw)
+                            # no fromfile(), see
+                            # https://github.com/numpy/numpy/issues/13319
+                            values[k, :] = np.frombuffer(raw,
```
tstenner: frombuffer doesn't need a character format string and is about 30% faster in one of my tests

pyxdf/pyxdf.py
cbrnr: Please don't use an OrderedDict if you don't really need it. `self.fmtidx` is never needed, because you should index with the dict's keys.

tstenner: In the chunk7 proposal I have one field for the LSL type index (which is an uint8), so in the future it is needed.

cbrnr: Why is this necessary? The format is already defined in the stream header, and having another place where this is stored is not ideal. In addition, I also don't like the idea that the index of a list is meaningful.

tstenner: It's one additional byte in the chunk header that has to be saved and processed. On the other hand, 

- each chunk is self-contained (all necessary informations like dimensions and data format to read the chunk data are in the chunk header),
- it allows additional level of error-checking by comparing the chunk header with the information in the stream header, either if the XDF writer changed the format mid-stream (unlikely) or the reader was left in the wrong position and continues to interpret stream data as chunk information.

The full type name would be better / more informative but it would either require a length and then the type name or a fixed length type name that might not be long enough if other types are added in the future.

cbrnr: XDF always consists of multiple chunks, so a chunk need not be self-contained. It's not about the additional byte in the chunk7 header, but we already have a format definition, and having it in two places will create maintenance problems at some point. Therefore, I'd remove this information from the chunk7 header, which will make this code unnecessary too.

tstenner: I still think it's a good idea, but unrelated to this PR so I have removed it.

cbrnr: Thanks, we can continue discussing this in the chunk7 proposal.

cbrnr:
We really need to set up a CI service. I don't have the time to verify if these changes (especially the switch from `struct.unpack` to `np.frombuffer`) did not introduce regressions, but by just looking at the code it LGTM. @cboulay feel free to do more checks and/or merge.

tstenner:
I created an [example file repository](https://github.com/xdf-modules/example-files) with a tiny (2kB) example file.

It's not a replacement for a CI, but the most basic test (`python -c "import pyxdf; print(pyxdf.load_xdf('../example-files/minimal.xdf'));"`) produces the sample output except for the changed dtype in the output time_series:

```
([{'info': defaultdict(<class 'list'>, {'name': ['SendDataC'], 'type': ['EEG'], 'channel_count': ['3'], 'nominal_srate': ['10'], 'channel_format': ['int16'], 'created_at': ['50942.723319709003'], 'desc': [None], 'uid': ['xdfwriter_11_int'], 'effective_srate': 10.000000000000021}), 'footer': {'info': defaultdict(<class 'list'>, {'writer': ['LabRecorder xdfwriter'], 'first_timestamp': ['5.1'], 'last_timestamp': ['5.9'], 'sample_count': ['9'], 'clock_offsets': [defaultdict(<class 'list'>, {'offset': [defaultdict(<class 'list'>, {'time': ['50979.76'], 'value': ['-.01']}), defaultdict(<class 'list'>, {'time': ['50979.86'], 'value': ['-.02']})]})]})}, 'time_series': array([[192, 255, 238],
       [ 12,  22,  32],
       [ 13,  23,  33],
       [ 14,  24,  34],
       [ 15,  25,  35],
       [ 12,  22,  32],
       [ 13,  23,  33],
       [ 14,  24,  34],
       [ 15,  25,  35]], dtype=int16), 'time_stamps': array([5. , 5.1, 5.2, 5.3, 5.4, 5.5, 5.6, 5.7, 5.8])}, {'info': defaultdict(<class 'list'>, {'name': ['SendDataString'], 'type': ['StringMarker'], 'channel_count': ['1'], 'nominal_srate': ['10'], 'channel_format': ['string'], 'created_at': ['50942.723319709003'], 'desc': [None], 'uid': ['xdfwriter_11_int'], 'effective_srate': 10.00000000000004}), 'footer': {'info': defaultdict(<class 'list'>, {'writer': ['LabRecorder xdfwriter'], 'first_timestamp': ['5.1'], 'last_timestamp': ['5.9'], 'sample_count': ['9'], 'clock_offsets': [defaultdict(<class 'list'>, {'offset': [defaultdict(<class 'list'>, {'time': ['50979.76'], 'value': ['-.01']}), defaultdict(<class 'list'>, {'time': ['50979.86'], 'value': ['-.02']})]})]})}, 'time_series': [['<?xml version="1.0"?><info><writer>LabRecorder xdfwriter</writer><first_timestamp>5.1</first_timestamp><last_timestamp>5.9</last_timestamp><sample_count>9</sample_count><clock_offsets><offset><time>50979.76</time><value>-.01</value></offset><offset><time>50979.86</time><value>-.02</value></offset></clock_offsets></info>'], ['Hello'], ['World'], ['from'], ['LSL'], ['Hello'], ['World'], ['from'], ['LSL']], 'time_stamps': array([5.1, 5.2, 5.3, 5.4, 5.5, 5.6, 5.7, 5.8, 5.9])}], {'info': defaultdict(<class 'list'>, {'version': ['1.0']})})
```

PR #3, opened 2019-04-16
Do not sort streams

cbrnr:
Sorting streams by name fails if two streams have the same name. To fix this, I've removed the sorting.


cboulay:
We'll deal with whether or not we return a dict / OrderedDictionary at another time. This quick fix shouldn't break anything.

PR #6, opened 2019-04-16
Fix _scan_forward (incorrect seek)

cbrnr:
As mentioned in https://github.com/sccn/xdf/issues/34, `_scan_forward` should advance 16 bytes and not 15.


cboulay:
I would love a test file for this but it seems correct.

cbrnr:
Yes, we should add tests. I can take care of some, but in this case I've checked with a hex editor and 16 is correct.

PR #7, opened 2019-04-16
Support pathlib objects

cbrnr:
Fixes https://github.com/sccn/xdf/issues/41.

I've also managed to shave off one level of precious indentation space in case you're wondering why the change is so big.

### DIFF
--- a/pyxdf/pyxdf.py
+++ b/pyxdf/pyxdf.py
@@ -12,6 +12,7 @@ import gzip
 import xml.etree.ElementTree as ET
 from collections import OrderedDict, defaultdict
 import logging
+from pathlib import Path
 
 import numpy as np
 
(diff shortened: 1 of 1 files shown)


cboulay:
The original `with <open> as f:` is preferred because it automatically closes the file when the context is left, either naturally at the end of loading or if there's an exception. In the PR form there is no f.close() so the file remains open.

cbrnr:
Of course I forgot the `close` 😄. I normally prefer the context manager too, but in this case I think it is better to explicitly call `open` and `close` because we really need that horizontal space.

tstenner:
What if an exception occurs somewhere before the `close`?

I'm all for reducing the indentation level, but it's not related to the pathlib changes and it interferes with a lot of other PRs.

cbrnr:
OK, I'll revert this, but we need to take care of exceptions anyway, because right now we catch all exceptions at the end. Technically, we could also add `f.close` to a `finally` block, but we can discuss this in a later PR.

tstenner:
Looks good to me. If @cboulay doesn't have any objections we should squash and merge it.

cboulay:
I merged a squashed version of this.

PR #10, opened 2019-04-18
Move example to pyxdf folder, add optional command line argument

tstenner:
Allow the example to be called from the installed module with an optional xdf file name, e.g. `python -m pyxdf.example` or `python -m pyxdf.example some_xdf_file.xdf`


cbrnr:
Thanks @tstenner!

PR #13, opened 2019-04-18
Move code out of function body for less indentation and easier cythonization

tstenner:
Screens have a limited width and Cython makes it easier to compile top level functions / classes so I moved the StreamData class and chunk 3 reader block out of the function body.

The next step would be a `pyxdf.pxd` file with annotations for Cython, leaving the [pure python](http://docs.cython.org/en/latest/src/tutorial/pure.html) code untouched.


cbrnr:
This is great! +1 for merge.

cbrnr:
Thanks @tstenner!

PR #17, opened 2019-04-28
Use Cython

tstenner:
As first approach to speed up the chunk 3 reading, I've added a pxd file so cython can use static types.

Usage: run `cythonize -3 -i -a pyxdf.py` in the pyxdf subfolder, import pyxdf as usual.

So far, it yields a speedup of about 16% (3.2s -> 2.7s), whereas my previous changes combined reduced the time from about 7s to 3.83s, so I'm not sure if it's worth it without replacing some functions with cython code.

**Edit**: It works with the python file classes, so gzip support is still there.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -105,3 +105,7 @@ venv.bak/
 
 # PyCharm
 .idea
+
+# cython temporaries
+/pyxdf/pyxdf.c
+/pyxdf/pyxdf.html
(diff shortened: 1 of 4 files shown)


tstenner:
This has become usable, has the same interface and almost no performance impact without the compiled extension (±.5%), so it could be merged.

cbrnr:
Very nice, that's a good start! Can you add a CHANGELOG entry?

tstenner:
Done.

cbrnr:
Thanks @tstenner!

PR #18, opened 2019-04-29
Initial version of CHANGELOG.md

### LINKED ISSUE #15: Create what's new
We should document all changes in a what's new file. I suggest that we create a Markdown file in the project root, but I'm not sure if there's a best practice regarding file name or format.

cbrnr:
Fixes #15. I included only stuff relevant for users, so internal code refactors are not in the list to keep it short. Let me know if you want to include such things, or if I forgot anything. Of course we can always edit this file later.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## [1.15.1] - 2019-04-26
 ### Added
 - Support pathlib objects (#7).
+- Allow example script to be called with an optional XDF file name (e.g. `python -m pyxdf.example` or `python -m pyxdf.example xdf_file.xdf`) (#10).
 
 ### Fixed
 - Use correct data types (#2).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,10 @@
+## [1.15.1]
+### Added
+- Support pathlib objects (#7).
+
+### Fixed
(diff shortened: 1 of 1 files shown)


tstenner:
I'd keep internal changes out of the changelog (they are in the git history for those who are _really_ interested). Some changelogs include the author name, do we also want to include them?

The added section is missing #10, other than that I think it covers the important changes.

cbrnr:
I've added #10. We should probably add authors, yes - would you add real names or GitHub handles?

tstenner:
Why not both, i.e. `[Name](https://github.com/githubhandle)`?

cboulay:
It's a little bit strange to include only user-facing changes in the CHANGELOG while at the same time providing named attribution to these changes. The result is that contributors that work on non user-facing things are excluded. I think it's better to just have a list of ALL contributors at the end, without specifying what they contributed.

cbrnr:
Good point. We could have a section for internal (not user-facing) changes. In the majority of cases, these will have noticable consequences such as improved speed etc. I'd rather list the contributer for each entry.

cbrnr:
Let's add all internal changes in a "Changes" section. I've already done that, plus I've added the names and links to PRs.

cbrnr:
Are you guys happy with the current format? If yes please merge.

PR #19, opened 2019-04-30
Include stream IDs in list of streams

cbrnr:
The only way to uniquely identify a stream is to use its stream ID. Therefore, we should include this information in the returned list of streams (inside the `info` key under a new `stream_id` key).

### DIFF
--- a/pyxdf/pyxdf.py
+++ b/pyxdf/pyxdf.py
@@ -346,6 +346,7 @@ def load_xdf(filename,
     for k in streams.keys():
         stream = streams[k]
         tmp = temp[k]
+        stream['info']['stream_id'] = k
         stream['info']['effective_srate'] = tmp.effective_srate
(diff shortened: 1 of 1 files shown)


cbrnr:
@cboulay or @tstenner I need this change in MNELAB, could you please merge if you are OK with it?

tstenner:
I have some lingering doubts about corner cases, but these won't be resolved by leaving out the streamid, so it's a +1 from me.

Could you also include a check if there's already a key named `stream_id`? It's unlikely, but it shouldn't be overwritten silently.

cbrnr:
What would you do if the key is already present? This can only be the case if someone explicitly writes a `stream_id` field in the `info` XML, which is not covered by the XDF specification. I think the most sensible thing to do in such a case is to issue a warning but still use the true stream ID found in the StreamHeader chunk.

PR #21, opened 2019-05-02
Fix loading long description in setup.py

cbrnr:
I think this needs to be changed because I get an error when I run `setup.py` locally. `README.md` is now in the same folder as `setup.py`, so going up a folder isn't correct. But I'm not 100% sure, please double-check.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,4 +1,4 @@
-"""Python setup script for the pylsl distribution package."""
+"""Python setup script for the pyxdf distribution package."""
 
 from setuptools import setup, find_packages
 from codecs import open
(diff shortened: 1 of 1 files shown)


cboulay:
It wasn’t going up a folder. I removed the .. when setting up the CI. 

I’m on mobile so I can’t check, but will this work if you run the command from a different folder (i.e., your working directory isn’t the root of pyxdf)?

That’s probably  not an important feature.

cbrnr:
OK, this explains why the PyPI package is working. I can revert the change and use the previous code without the `'..'`, that way this will work for any working directory.

cbrnr:
I forgot to rebase and didn't see that this change was already in master. So this only contains minor cosmetic changes, so I went ahead and merged it.

PR #23, opened 2019-05-15
Sort streams by stream_id

cbrnr:
Since we now store the unique stream IDs, we can just as well sort the list of streams now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## [Unreleased]
 ### Added
 - Store unique stream ID inside the `["info"]["stream_id"]` dict value ([#19](https://github.com/xdf-modules/xdf-Python/pull/19) by [Clemens Brunner](https://github.com/cbrnr)).
+- Sort list of streams by stream ID ([#23](https://github.com/xdf-modules/xdf-Python/pull/23) by [Clemens Brunner](https://github.com/cbrnr))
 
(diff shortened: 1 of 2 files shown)


cbrnr:
@tstenner feel free to merge (or let me know if you don't like this change).

tstenner:
What's the benefit? As far as I can recall, the stream IDs are based on the order LabRecorder resolves the streams so it's neither meaningful nor identical between multiple recording sessions, even for the same streams.

cbrnr:
Correct, the only benefit is that I don't have to sort the list when I display a table of all streams in MNELAB, which just looks prettier. I can of course do it there, but I thought since it doesn't hurt it might as well be done here (the computational overhead is minimal in any case).

cbrnr:
Let's not sort then, if I need it I can do it myself.

tstenner:
Wouldn't another criterion make more sense for MNELAB (e.g. type+name)? Ordering by the streamid replaces a somewhat arbitrary ordering with another arbitrary ordering

cbrnr:
Yes, the table I display is sortable by any column (by clicking on the corresponding header). I just thought that by default I'll sort by ID. I get your point, but within one recording session the stream IDs are the only unique identifiers of a stream, that's why this is more or less the only sensible default sort option IMO.
Screen Shot 2019-05-16 at 11 44 16

PR #26, opened 2019-05-22
License

### LINKED ISSUE #25: Licensing
The main module `pyxdf.py` currently contains the following copyright statement at the top of the file:
```
Copyright (c) 2015-2018, Syntrogi Inc. dba Intheon
```

The docstring of `load_xdf` contains another copyright notice:
```
    License:
        This file is covered by the BSD license.

        Copyright (c) 2015-2018, Syntrogi Inc. dba Intheon

        Redistribution and use in source and binary forms, with or without
        modification, are permitted provided that the following conditions are
        met:

            * Redistributions of source code must retain the above copyright
              notice, this list of conditions and the following disclaimer.
            * Redistributions in binary form must reproduce the above copyright
              notice, this list of conditions and the following disclaimer in
              the documentation and/or other materials provided with the
              distribution

        THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
        "AS IS AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
        LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
        A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
        OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
        SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
        LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
        DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
        THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (
        INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
        OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
```

How do we proceed? First, we need to update the year to 2019. Second, I am starting to change more in the code so it would be nice if this was acknowledged somewhere. Third, I think we should put the current maintainers as copyright holders (but of course continue to mention Syntrogi Inc. dba Intheon). Here's a suggestion of what I'd put at the top of the file:
```
Authors: Syntrogi Inc. dba Intheon
         Clemens Brunner <mail address>
         Tristan Stenner <mail address>
         Chadwick Boulay <mail address>
         <others?>

License: BSD (2-clause)
```
Then I'd remove the copyright and license notice from the function docstring completely.

It would be even nicer if we could add the names of the original authors instead of/in addition to Syntrogi (@chkothe?).

Thoughts @cboulay @tstenner?

cbrnr:
Fixes #25.

I think Syntrogi/Intheon should/must accept this change to replace the copyright holder "Syntrogi Inc. dba Intheon" with "XDF-Python Development Team" before we can merge this.

@trmullen are you OK with this change? Also, would you like to have the names of the contributors instead of Syntrogi in the authors list at the top of the file? I'd prefer to have @chkothe and anyone else who contributed from Syntrogi instead of the company name.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,25 +1,28 @@
 BSD 2-Clause License
 
-Copyright (c) 2015-2018, Syntrogi Inc. dba Intheon
+Copyright (c) 2015-2019, XDF-Python Development Team
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
 modification, are permitted provided that the following conditions are met:
 
-* Redistributions of source code must retain the above copyright notice, this
-  list of conditions and the following disclaimer.
+1. Redistributions of source code must retain the above copyright notice, this
+   list of conditions and the following disclaimer.
+2. Redistributions in binary form must reproduce the above copyright notice,
+   this list of conditions and the following disclaimer in the documentation
+   and/or other materials provided with the distribution.
 
-* Redistributions in binary form must reproduce the above copyright notice,
-  this list of conditions and the following disclaimer in the documentation
-  and/or other materials provided with the distribution.
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+(INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
-AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
-IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
-DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
-FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
-DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
-SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
-CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
-OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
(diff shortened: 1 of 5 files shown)


cbrnr:
Which authors should I add for `__init__.py`, `example.py`, and `setup.py`? I think these are all files we currently have.

cboulay:
Christian developed pyxdf as part of his duties at Intheon. 
I think Alejandro Ojeda also contributed to its development early on.

I have no expertise w.r.t. copyright law and it would be dangerous for me to suggest changing/not changing without any knowledge of the implications, and for that reason I tend to "leave it as it is".

cbrnr:
I think it's safe to make the changes related to formatting (i.e. remove the license from the docstring because we already have it in our main folder, make a short header with contributors at the top of each file) as long as we do not change Syntrogi. If we want to make this change as well we need explicit consent from them and we're good to go.

chkothe:
As for the suggested edits to the root license file, pyxdf was developed by our team at Intheon and we're continuing to maintain it (and we are building more on top of it). We would like the original copyright notice in the root license to be retained. I don't think the suggested amendment about the views of the authors is necessary, and it's not in OSI-approved BSD license text, I'd say we keep the text.

About the other copyright notices in `pyxdf.py,` I agree that these notices are redundant and complicate the licensing situation, and I'm fine with stripping them out.

Regarding a contributors list at the top of the source file(s) -- having such a list there is a good change and allows individual contributors or their organizations to be acknowledged. At Intheon, we usually attribute the team and not our individual team members though, so we would likely not at this point replace Intheon by a list of individuals.

I think that going forward, if organizations and/or individuals make major contributions, they could also propose to have a copyright line added to the root license file, which could then be decided on a case-by-case basis (e.g., changing a comma in a code comment would be unlikely to qualify for that, while the person may of course still append themselves to the respective file's contributors list).

@trmullen any thoughts on the proposal?

trmullen:
Hi @cbrnr, I agree with @chkothe’s points. Since pyxdf was written at Intheon, the institutional copyright notice should be retained per BSD and standard OSI practice. But of course this copyright has no bearing on others' rights to incorporate or extend the pyxdf module under the terms of the BSD license, and add to the license file whatever copyright / attribution they wish for their own contributions. Perhaps this modification to the root License file could be done as part of a contributor's PR.

On a more technical note, I’m not sure that “XDF-Python Development Team” can actually hold copyright since it's not an entity nor a clearly defined group of individuals or entities. When an OS project has many authors, it's not uncommon to list copyright as held by "the authors of <the software name>”. But in the case of xdf-Python, since there are just a few contributing individuals/institutions — and in some cases the “author” may not actually be the copyright holder —  it seems most straightforward to append each contributing institution's (or individual's, where appropriate) preferred copyright line in the license file e.g. similar to http://www.fieldtriptoolbox.org/copyrights/

Regarding the contributor list, I agree with @chkothe’s proposal there. As he pointed out, in our group, we usually acknowledge efforts as a team. However, given @chkothe's central role in developing pyxdf and, more generally, XDF, I would want to specifically acknowledge him by adding him to the contributor list as such:

Contributors:
Christian Kothe & the Intheon pyxdf team
<others>

cheers —

cbrnr:
Thanks for your input. I'll go ahead and update this PR accordingly.

>  I don't think the suggested amendment about the views of the authors is necessary, and it's not in OSI-approved BSD license text, I'd say we keep the text.

I took the text from [Wikipedia](https://en.wikipedia.org/wiki/BSD_licenses#2-clause_license_(%22Simplified_BSD_License%22_or_%22FreeBSD_License%22)), which includes this paragraph. But I'm fine with keeping the text as it was because there are no further differences.

Amending the root copyright file with (major) contributors is a great idea. According to the log, @cboulay, @tstenner, and @cbrnr have made contributions that exceed changing a comma, so I'd go ahead and add these three persons to the root license.

> On a more technical note, I’m not sure that “XDF-Python Development Team” can actually hold copyright since it's not an entity nor a clearly defined group of individuals or entities

I'm not a legal expert on OS licenses, but the XDF-Python group is clearly defined at least as a GitHub group. Which of course does not mean it is a legal entity, but since I don't want to get into that kind of stuff I think your proposed solution to add individuals to the license works around this problem (although I had hoped that we could simplify things by just mentioning the group).

cbrnr:
Can you please review the changes and let me know if you're happy? Currently, `setup.py` lists @chkothe's Gmail address, I don't know if you want to change this (let me know).

chkothe:
Good catch re the Author email, that should indeed be christian.kothe@intheon.io. More importantly though, can you please retain the "Syntrogi Inc. dba Intheon" in the copyright line? But thanks for updating the year. Tim also noted that there's one more thing in the license text that deviates from today's BSD license style, the "All rights reserved." clause (which was there for historic reasons but is [omitted](https://www.freebsd.org/internal/software-license.html) for newer licenses). It would be better to remove it, otherwise it may cause unnecessary panic when a potential XDF user checks the text against the OSI template.  Other than that, looks good!

cbrnr:
Done! This should be ready for merge now. @cboulay or @tstenner please go ahead and merge if you are happy with the changes.

PR #30, opened 2019-06-07
Update metadata in setup.py

### LINKED ISSUE #29: Update PyPI information
I noticed that several fields at the [PyPI website](https://pypi.org/project/pyxdf/) still point to the old location or don't work as expected.

- Homepage should link to  https://github.com/xdf-modules/xdf-Python
- Author currently shows "('Christian Kothe, Tristan Stenner', 'Clemens Brunner')" as one links to Christian's email. I don't know if it is possible to specify multiple authors. I've checked some projects with multiple authors (e.g. pandas, scipy), and they omit the author field completely. We should consider doing the same, and instead of using the author field populate the maintainers list (which supports multiple persons).
- The project description should mirror README.md from our repository (and BTW this should be updated because e.g. uploading to PyPI is now much easier).

cbrnr:
Fixes #29. @cboulay please merge if you're happy. I left Christian in the author field, but we should really think about having a general mail address that all maintainers can use (e.g. pyxdf@googlegroups.com or something like that) and put that address in the maintainer email field instead.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,8 +1,7 @@
 [metadata]
-# This includes the license file in the wheel.
+# includes the license file in the wheel
 license_file = LICENSE
 
 [bdist_wheel]
-# This flag says that the code is written to work on both Python 2 and Python
-# 3.
+# our code is written to work on both Python 2 and 3
 universal=1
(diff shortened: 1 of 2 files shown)


cboulay:
OK it looks good.

BTW, if you still have the commands handy, please update the README 'for maintainers' part with instructions for tagging with a version number and pushing to trigger an update on pypi.

Issue #34, opened 2019-09-11
Critical Bug in _jitter_removal


ollie-d:
Hello,

Please forgive me if I am posting this in an undesired format. I simply want to explain the bug, how I found it, and what the solution is.

**The bug**
  When jitter removal is enabled, the current code creates an invalid mapping, thus distorting time. I am not sure if the effect is as dramatic as what I have experienced, but in my case it renders my markers completely useless.

**Problematic code** (line 563 of pyxdf.py):
`indices = np.hstack((0, indices, indices, nsamples - 1))`

**How I Found it**
I collected some pilot data and realized that my classifier was at chance. After going through all of the relevant code, I determined that my filtering and classifying were being conducted correctly. I then ran my analysis on simulated data, and the result was perfect. I then started analyzing my code for epoching the data. Luckily, I have started collecting photosensor channels so as not to rely too heavily on software markers. When I examined the markers position relative to the photosensor, I realized they were not aligned:
![image](https://user-images.githubusercontent.com/23530226/64742191-3466dd80-d4b0-11e9-9625-de275747ec5d.png)
I then decided to look at my data in MATLAB via the load_xdf() function shipped with EEGLAB, and my data looked as expected:
![image](https://user-images.githubusercontent.com/23530226/64742254-69733000-d4b0-11e9-8366-7aa6679f95f1.png)
After that, I decided to individually change all of the options in python and discovered that when dejitter_timestamps was set to False, my data looked as expected:
![image](https://user-images.githubusercontent.com/23530226/64742290-90316680-d4b0-11e9-87f4-4782071cfe9b.png)

After determining that the removing of jitter was the problem, I went through each line of the relevant code in MATLAB and python and discovered that the ranges being used were inconsistent. For example, with my data, MATLAB first finds a problem in the range [2 11]. I expected to see that python would show the same problem in [1 10], but instead it finds a problem at [0 10]. I noticed that every index that python found had a lower bound by 1 index compared to MATLAB, even when taking 1-indexing into account:
![image](https://user-images.githubusercontent.com/23530226/64742512-5745c180-d4b1-11e9-9724-2d4866c8dfa1.png)
This subsequently caused a discrepancy in the mapping values between python and MATLAB.

**Solution**
Increment the first index by 1 to match MATLAB. 
`indices = np.hstack((0, indices+1, indices, nsamples - 1))`

**Conclusion**
[...]

PR #35, opened 2019-09-17
_jitter_removal Fix

ollie-d:
Modified '_jitter_removal' in order to match the MATLAB equivalent function

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 ## [Unreleased] - XXXX-XX-XX
 ### Added
 - Add option to load only specific streams ([#24](https://github.com/xdf-modules/xdf-Python/pull/24) by [Clemens Brunner](https://github.com/cbrnr)).
+### Fixed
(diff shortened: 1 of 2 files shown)


cbrnr:
Could you please add an entry to `CHANGELOG.md` (in the "Fixed" section which you have to create)?

ollie-d:
Sorry, I wasn't sure if I should make the entry myself. Please let me know if there's anything else I should do!

cbrnr:
Thanks, I pushed a minor change (we don't need to mention that it now matches what the MATLAB implementation does, because all that counts is that the bug is now fixed).

I'll wait a bit to see if @tstenner or @cboulay have time to take a look, but if not I'd really like to merge this soon (and push a new release).

cboulay:
I’m in transit now. I’ll take a look tomorrow.

cboulay:
For #34

cboulay:
I'm going to merge this in then make some additional changes. Briefly, I think the `np.hstack((0, indices + 1, indices, nsamples - 1))` line and the following reshaping line is a little bit hard to decipher with a quick scan. I'm going to be very explicit about creating a vector of range_starts and range_stops.

PR #37, opened 2019-09-20
Pass string instead of Path object to open

hankso:
## What is the error
A path object cannot be directly opened like a filename string.

## How to reproduce
```python
>>> from pathlib import Path
>>> p = Path('xdf_sample.xdf')    # it's a path object
>>> open(Path('xdf_sample.xdf'))  # but open only accept str / bytes
TypeError: invalid file: PosixPath('xdf_sample.xdf')
```
Same as `gzip.open`
```
>>> import gzip
>>> gzip.open(Path('foo'))
TypeError: filename must be a str or bytes object, or a file
``` 

## How to fix
```python
>>> p = Path('foo')
>>> open(str(p))
```
---
BTW, I suggest NOT to mix using `os.path` and `pathlib`. So all `os.path` functions are replaced.

Please consider merging.

> Python 3.5.2
> Ubuntu 16.04

### DIFF
--- a/pyxdf/pyxdf.py
+++ b/pyxdf/pyxdf.py
@@ -10,7 +10,6 @@
 This function is closely following the load_xdf reference implementation.
 """
 
-import os
 import struct
 import itertools
 import gzip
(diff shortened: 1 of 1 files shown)


cboulay:
This is a 3.5 backwards compatibility patch. In Python 3.6 and up, `open` accepts Pathlike objects.
It looks good to me. Thanks for the PR!

cbrnr:
@hankso would you like your real name to be shown in the changelog for your contribution?

PR #38, opened 2019-09-20
pytest with datafile

cboulay:
This is a first attempt. I would like to NOT include the datafile in the repo and instead download the file on demand. We can put the (small) datafile in a GitHub release, otherwise we need to host them somewhere.

@tstenner , how do you think we should get azure to download our test data? Within a script? Or within a pytest fixture? Something else?

### DIFF
--- a/pyxdf/test/test_library_basic.py
+++ b/pyxdf/test/test_library_basic.py
@@ -2,6 +2,7 @@ import pyxdf
 import pyxdf.pyxdf
 import pytest
 import io
+import os
 
 
 #%% test
@@ -24,3 +25,31 @@ def test_read_varlen_int():
     assert vla(b'\x08\xfd\x12\x00\x34\x12\x34\x56\x78') == 0x78563412340012fd
     with pytest.raises(RuntimeError):
         vla(b'\x00')
+
+
+FIXTURE_DIR = os.path.join(os.path.dirname(os.path.realpath(__file__)), 'test_files')
+
+
+@pytest.mark.datafiles(os.path.join(FIXTURE_DIR, 'xdf_sample.xdf'))
+def test_load_sample(datafiles):
+    for xdf_file in datafiles.listdir():
+        streams, fileheader = pyxdf.load_xdf(xdf_file)
+
+        stream_names = []
+        for strm in streams:
+            assert('info' in strm)
+            assert('name' in strm['info'])
+            assert(isinstance(strm['info']['name'], list))
+            stream_names.append(strm['info']['name'][0])
+            assert(stream_names[-1] in ['MousePosition', 'MouseButtons',
+                                        'Keyboard', 'AudioCaptureWin'])
+
+            assert('time_series' in strm)
+            import numpy as np
+            assert(isinstance(strm['time_series'], (np.ndarray, list)))
+            assert('time_stamps' in strm)
+            assert(isinstance(strm['time_stamps'], np.ndarray))
+            if isinstance(strm['time_series'], np.ndarray):
+                assert(strm['time_stamps'].shape[0] == strm['time_series'].shape[0])
+            else:
+                assert(strm['time_stamps'].shape[0] == len(strm['time_series']))
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -34,7 +34,7 @@ steps:
   displayName: 'Install Dependencies'
 
 - script: |
-    pip install pytest pytest-azurepipelines
+    pip install pytest pytest-azurepipelines pytest-datafiles
     pytest
   displayName: 'pytest'
 


pyxdf/test/test_library_basic.py
```diff
@@ -24,3 +25,31 @@ def vla(data: bytes):
     assert vla(b'\x08\xfd\x12\x00\x34\x12\x34\x56\x78') == 0x78563412340012fd
     with pytest.raises(RuntimeError):
         vla(b'\x00')
+
+
+FIXTURE_DIR = os.path.join(os.path.dirname(os.path.realpath(__file__)), 'test_files')
```
cbrnr: Can you do this with pathlib?

cbrnr:
Would git-lfs or git-annex be a viable option? Why don't you want to include (small) test files directly in this repo? If they are small, I don't see the problem (the test files should not be packaged of course).

In MNE, we have a separate repository just for our testing data, but the main problem is that downloading is super slow. Maybe this is faster when you attach files to a release, but isn't this difficult to manage? Do we have to move/copy all testing files for each new release then?

pyxdf/test/test_library_basic.py
cboulay: I was just following the most straightforward example from pytest-datafiles. These small details aren't important until we iron out the bigger details, which may require using a completely different way of downloading/loading data.

cboulay:
> Why don't you want to include (small) test files directly in this repo?

(1) The way the package is structured currently, the tests travel with the python package to pypi. We could probably put the test data in a different folder, or manually remove them from the packages before uploading (seems like silly over-engineering to do it that way). This is also exactly the same test file that's in the parent repo.
(2) I don't want to inflate the git history with binaries. Seems not so bad right now, but it can become a problem quickly, especially because...
(3) there are plans (SCCN/Intheon) to add many more test files to the test suite, especially 'problematic' files.

Attaching to a release feels like the wrong solution. The main reason I want the test files in there at all is to automatically test file loading whenever someone makes a PR. If a new feature requires a new test file, how does one verify the feature before the test file is in a release?

cbrnr:
I totally agree that we need a test suite that automatically runs with a PR. The more files we have in there the better. I think git-lfs should work pretty well for our purposes.

tstenner:
> I would like to NOT include the datafile in the repo and instead download the file on demand.
Seconded. It's not always practical to clone a shallow copy and processing any data file every present in the repository slows everything down (or, as Clemens put it: "super slow") . With Git LFS it would be reasonable, but I think it's overkill. Also, once a big file gets pushed it's in the repository practically forever.

Also, once we isolate the broken part of a file it should be possible to create a synthetic test file that triggers the bug, but is much shorter.

For the benchmarks, we need huge files so I have small scripts that generate big files to test the different code paths on the fly and then delete them after. Those could stay in the repository.

> @tstenner , how do you think we should get azure to download our test data?

It all depends what the CI should test. For one job (i.e., only one operating system), it doesn't really matter. For parallel jobs on Azure, I'd create one stage to download the files and then copy them to each job so the files get downloaded (and possibly cached) once per run and copying them to the CI workers is fast.

> Within a script? Or within a pytest fixture? Something else?

I'd suggest in a script unless pytest handles downloading changed/new files efficiently. rsync would do this, but it'd need a dedicated host. Or a simple folder listing served via HTTP and wget.

cboulay:
> but it'd still be nice to also test with real-world large files

Not with CI systems though. If you want to include some tests in the repo that are only run on your local computer (e.g., after checking for the presence of a specific folder and/or only invoked with a special argument to pytest) then that's fine. If you want the tests to be run on the CI servers then the files should be small or generated on the fly.

cbrnr:
Yes, obviously we can't test large files with CIs, so I'm fine with only using small files for CIs. We should still try to run our tests on large files, because these might reveal issues that are not found in small files. We could run our tests on all files (small and large) manually before a new release. I'm open to the implementation details of how we want to make this work, and we also still need to solve the question where we store our large files (since I assume that you all seem to be OK with including small files in either this repo or a separate xdf-testdata repo - I'd prefer the separate repo because that way other repos can re-use the same test files).

cboulay:
Agreed, not within this repo. I don't have enough experience with git-lfs, does it have better performance with large binary files than normal git?

tstenner:
> Does it have better performance with large binary files than normal git?

It basically uploads the files separately and replaces them with links, so it's a lot faster than normal git but only as fast as the place where the files get uploaded to.

cbrnr:
Exactly. And we still don't know where to host the large files.

cbrnr:
@cboulay could you go ahead and create an `xdf-testdata` repo within the `xdf-modules` organization?

cbrnr:
Re how to do the actual testing, I think `pytest-datafiles` is not even required. This package's main purpose is to copy files to a temp dir in order to avoid modifying the original files. We don't need to do this, because (1) we're pulling our test files from a git repo (and we don't push), and (2) we don't modify our test files anyway.

I'd clone the example-files repo in an Azure script, and we can then use the test files with the `@pytest.mark.parametrize` decorator.

PR #39, opened 2019-09-20
Add badges to README.md

cbrnr:
I thought that these colorful badges look nice.


cbrnr:
Preview: https://github.com/xdf-modules/xdf-python/blob/595428e3d32a6ef669192233b34d025578d45ebf/README.md

cboulay:
The generated version and license badges are great, thank you.

But the 3.5/6/7 badges are artificial and require manual maintenance so I would prefer not to include those.

cbrnr:
What about including a "python 3" badge then? This should be valid for the foreseeable future.

cbrnr:
I'll take care of keeping the badges in sync. If at some point you decide you don't want them feel free to remove them.

PR #42, opened 2019-09-25
Restore verbose kwarg to load_xdf

### LINKED ISSUE #41: Re-add verbose argument to `load_xdf`
The short of it is that I updated some Intheon code to no longer use 'verbose' in the call to `load_xdf`. Then @chkothe reverted that changed and pinned their use of pyxdf to an older version. So obviously this is important to them though I haven't had a conversation as to why.

Looking back, it seems that I'm the one that removed it [in this commit](https://github.com/sccn/xdf/commit/0428fe0aaa5f8565141342c98b3610e6f9d5ed55#diff-8dbbae27ec1074622bb151a61e0fb80e). The only thing related I see is [this comment](https://github.com/sccn/xdf/pull/6#issuecomment-263812175). So I guess my interpretation was "why use a 'verbose' argument when you should be setting the logger level separately?".

So if we re-add the verbose argument, does that mean we should use it to override the logger level? Or if verbose is set, do we skip the logger output and print to screen?

cboulay:
Fixes #41

### DIFF
--- a/pyxdf/pyxdf.py
+++ b/pyxdf/pyxdf.py
@@ -77,7 +77,8 @@ def load_xdf(filename,
              clock_reset_threshold_stds=5,
              clock_reset_threshold_offset_seconds=1,
              clock_reset_threshold_offset_stds=10,
-             winsor_threshold=0.0001):
+             winsor_threshold=0.0001,
+             verbose=None):
     """Import an XDF file.
 
     This is an importer for multi-stream XDF (Extensible Data Format)
@@ -107,6 +108,10 @@ def load_xdf(filename,
             select_streams=[{'type': 'EEG'}, {'name': 'TestAMP'}], streams
             matching either the type *or* the name will be loaded.
           - None: load all streams (default).
+
+        verbose: Passing True will set logging level to DEBUG,
+          False will set logging level to WARNING,
+          and None will use root logger level. (default: None)
 
         synchronize_clocks : Whether to enable clock synchronization based on
           ClockOffset chunks. (default: true)
@@ -195,6 +200,8 @@ def load_xdf(filename,
         load the streams contained in a given XDF file
         >>> streams, fileheader = load_xdf('myrecording.xdf')
     """
+    if verbose is not None:
+        logger.setLevel(logging.DEBUG if verbose else logging.WARNING)
 
     logger.info('Importing XDF file %s...' % filename)
     filename = Path(filename).resolve()  # absolute path following symlinks
@@ -236,8 +243,8 @@ def load_xdf(filename,
                 chunklen = _read_varlen_int(f)
             except Exception:
                 if f.tell() < filesize - 1024:
-                    logger.warn('got zero-length chunk, scanning forward to '
-                                'next boundary chunk.')
+                    logger.warning('got zero-length chunk, scanning forward to '
+                                   'next boundary chunk.')
                     _scan_forward(f)
                     continue
                 else:
@@ -280,7 +287,6 @@ def load_xdf(filename,
                 # noinspection PyBroadException
                 try:
                     nsamples, stamps, values = _read_chunk3(f, temp[StreamId])
-
                     logger.debug('  reading [%s,%s]' % (temp[StreamId].nchns,
                                                             nsamples))
                     # optionally send through the on_chunk function


tstenner:
I'm not a fan of those `if verbose: log.debug()` calls. What about setting the default to `None` and if it's set changing the log level at the beginning and restoring it at the end?

cbrnr:
+100 for doing it like @tstenner suggested. I'd implement the following behavior:

- If `verbose=None` (should be the default), we don't change the logging level and use whatever is currently set by the user.
- If `verbose=True`, we set the logging level to `INFO`.
- If `verbose=False`, we set the logging level to `WARNING`.
- In addition, string arguments are also possible, e.g. `verbose=DEBUG' sets the logging level to `DEBUG`. These string arguments are equivalent to passing `logging.DEBUG` etc.

MNE does it like that, see https://www.nmr.mgh.harvard.edu/mne/stable/generated/mne.set_log_level.html.

cboulay:
I find it strange that INFO, which is technically above DEBUG, also includes DEBUG. Or am I getting that wrong?
Anyway, I force pushed the suggested changes. I think this is correct.

cbrnr:
No, INFO does not include DEBUG. It's the other way round, DEBUG includes everything else.

pyxdf/pyxdf.py
```diff
@@ -195,6 +200,8 @@ def load_xdf(filename,
         load the streams contained in a given XDF file
         >>> streams, fileheader = load_xdf('myrecording.xdf')
     """
+    if verbose is not None:
+        logger.setLevel(logging.INFO if verbose else logging.WARNING)
```
cbrnr: I think we should store the current log level and restore it at the end of the function. Maybe we could even write a `verbose` decorator for that purpose. This decorator would do all the context management and add a `verbose` argument to the function it wraps.

This is also my second comment - even if you do not want to implement that as a decorator, I think the `verbose` argument should be the last parameter of `load_xdf`.

tstenner: Or in `**kwargs`

cbrnr: Yep.

cbrnr: Example for decorator: https://www.nmr.mgh.harvard.edu/mne/dev/generated/mne.verbose.html

cboulay: Nope. I don't want to handle kwargs unless we're passing them on to something. There's no need for that here.

We also don't need a decorator for something that appears in exactly one function.

Please, less over-engineering.

cboulay:
> No, INFO does not include DEBUG. It's the other way round, DEBUG includes everything else.

That's what I guessed from reading about logging levels in the docs though it was never explicit, but then I read something that stated the opposite explicitly. So, if DEBUG is the lowest-level, then why did you recommend that when verbose is provided we set it to INFO instead of DEBUG? Is it just a case of "that's common practice"?

> I think we should store the current log level and restore it at the end of the function.

Why? What's the use case? This logger is for this module only.

As far as I can tell, you're trying to prevent a situation where someone first calls load_xdf(verbose=False), then later call load_xdf(verbose=None) expecting the logger to use the root logging level, not the logging level that was set from the previous call.

If someone ever complains about that then by all means you can make a decorator. Decorate your heart out.

cbrnr:
> That's what I guessed from reading about logging levels in the docs though it was never explicit, but then I read something that stated the opposite explicitly. So, if DEBUG is the lowest-level, then why did you recommend that when verbose is provided we set it to INFO instead of DEBUG? Is it just a case of "that's common practice"?

Yep.

> Why? What's the use case? This logger is for this module only.
>
> As far as I can tell, you're trying to prevent a situation where someone first calls load_xdf(verbose=False), then later call load_xdf(verbose=None) expecting the logger to use the root logging level, not the logging level that was set from the previous call.
>
> If someone ever complains about that then by all means you can make a decorator. Decorate your heart out.

Fair enough.

cbrnr:
Feel free to change the behavior of `verbose=True` to setting the log level to DEBUG. I think that's what @chkothe had in mind in the first place. Other than that, +1 for merge.

cboulay:
OK back to DEBUG. I'll run a couple more tests locally then merge.

PR #45, opened 2019-09-27
Remove Python 2 from setup.py

### LINKED ISSUE #44: Version 1.16.0 does not work with Python 2.7 anymore
```python
$ python -V
Python 2.7.15
$ python -c 'import pyxdf'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/Users/hoechenberger/miniconda3/envs/pyxdf/lib/python2.7/site-packages/pyxdf/__init__.py", line 11, in <module>
    from .pyxdf import load_xdf, resolve_streams, match_streaminfos
  File "/Users/hoechenberger/miniconda3/envs/pyxdf/lib/python2.7/site-packages/pyxdf/pyxdf.py", line 746
    chunk = {**chunk, **_parse_streamheader(xml)}
              ^
SyntaxError: invalid syntax
```

Metadata in `setup.py` says it should be compatible with both Py2 and Py3:
https://github.com/xdf-modules/xdf-python/blob/049aef3486040851f53ca710cee761c796b6835f/setup.py#L63-L64

cbrnr:
Fixes #44.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## [1.16.1] - 2019-09-28
+### Fixed
+- Remove Python 2 compatibility from `setup.py` ([#45](https://github.com/xdf-modules/xdf-Python/pull/45) by [Clemens Brunner](https://github.com/cbrnr)).
+
 ## [1.16.0] - 2019-09-27
 ### Added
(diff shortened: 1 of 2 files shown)


cboulay:
Let's get this in fast. You'll need a new version number and tag number for it to go to pypi.

PR #46, opened 2019-10-04
Preallocate the read buffer for numerical data

tstenner:
For numerical chunks, each sample has a constant size, so this change allocates and reuses a read buffer to avoid one allocation per sample. With a large sample file (64 double channels, 1M samples) the total runtime decreases by 11% and the context switches by 98%.


cbrnr:
Nice change!

cbrnr:
Can you add a CHANGELOG entry?

tstenner:
Done.

cbrnr:
Thanks @tstenner!

PR #47, opened 2019-10-07
Don't check srate for irregularly sampled streams

cbrnr:
If `stream.srate` is zero, we divided by zero in the effective vs. nominal srate check. This check should only be run for regularly sampled streams (i.e. with `stream.srate` > 0).

### DIFF
--- a/pyxdf/pyxdf.py
+++ b/pyxdf/pyxdf.py
@@ -592,11 +592,11 @@ def _jitter_removal(streams, threshold_seconds=1, threshold_samples=500):
                               stream.time_stamps[seg_starts])
                 stream.effective_srate = np.sum(counts) / np.sum(durations)
 
-        if np.abs(stream.srate - stream.effective_srate) / stream.srate > 0.1:
+        srate, effective_srate = stream.srate, stream.effective_srate
+        if srate != 0 and np.abs(srate - effective_srate) / srate > 0.1:
             msg = ("Stream {}: Calculated effective sampling rate {:.4f} Hz is"
                    " different from specified rate {} Hz.")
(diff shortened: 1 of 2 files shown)


pyxdf/pyxdf.py
```diff
@@ -592,11 +592,11 @@ def _jitter_removal(streams, threshold_seconds=1, threshold_samples=500):
...
                 stream.effective_srate = np.sum(counts) / np.sum(durations)
 
-        if np.abs(stream.srate - stream.effective_srate) / stream.srate > 0.1:
+        srate, effective_srate = stream.srate, stream.effective_srate
+        if srate != 0 and np.abs(srate - effective_srate) / srate > 0.1:
             msg = ("Stream {}: Calculated effective sampling rate {:.4f} Hz is"
                    " different from specified rate {} Hz.")
-            logger.warning(msg.format(stream_id, stream.effective_srate,
-                                      stream.srate))
+            logger.warning(msg.format(stream_id, effective_srate, srate))
```
tstenner: should  be `logger.warning(msg, effective_srate, srate)`

cbrnr: Why? `msg` is a string with `{}`, which need to be filled with `.format`.

cbrnr: I see, according to the documentation this should be possible, but it only works with a old-style `%` placeholders.

cbrnr: I'm using `%` now.

cbrnr:
@tstenner feel free to merge if you're happy. I have another PR incoming that fixes more warnings, but I'd like to wait until this gets merged.

PR #49, opened 2019-10-15
Add tests for example data to CI

cbrnr:
WIP: trying to implement testing using data in the example-files repo. Closes #38.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -31,16 +31,22 @@ steps:
 - script: |
     python -m pip install --upgrade pip wheel twine setuptools_scm
     pip install numpy
-  displayName: 'Install Dependencies'
+  displayName: 'Install dependencies'
+
+- script: |
+    git clone https://github.com/xdf-modules/example-files.git
+    pwd
+    ls
+  displayName: 'Download test data'
 
 - script: |
     pip install pytest pytest-azurepipelines
     pytest
-  displayName: 'pytest'
+  displayName: 'Test'
 
 - script: |
     python setup.py sdist bdist_wheel
-  displayName: 'Build Wheel'
+  displayName: 'Build wheel'
 
 - task: TwineAuthenticate@0
   condition: and(succeeded(), contains(variables['Build.SourceBranch'], 'refs/tags/'))
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -35,8 +35,6 @@ steps:
 
 - script: |
     git clone https://github.com/xdf-modules/example-files.git
-    pwd
-    ls
   displayName: 'Download test data'
 
 - script: |
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -34,7 +34,7 @@ steps:
   displayName: 'Install dependencies'
 
 - script: |
-    git clone https://github.com/xdf-modules/example-files.git
+    git clone --depth 1 https://github.com/xdf-modules/example-files.git
   displayName: 'Download test data'
 
 - script: |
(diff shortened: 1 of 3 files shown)


cbrnr:
@cboulay @tstenner tests on Azure work. We can now start adding more (small) test files to our example-files repo and they will be automatically smoke-tested (specific tests regarding the file contents need to be added manually).

cbrnr:
@tstenner could you review this PR and merge if you're happy? We have some nice changes already and I'd like to release 1.16.2 as soon as this gets merged.

azure-pipelines.yml
```diff
@@ -31,16 +31,20 @@ steps:
 - script: |
     python -m pip install --upgrade pip wheel twine setuptools_scm
     pip install numpy
-  displayName: 'Install Dependencies'
+  displayName: 'Install dependencies'
+
+- script: |
+    git clone https://github.com/xdf-modules/example-files.git
```
tstenner: This should be a shallow clone, or simply fetch https://github.com/xdf-modules/example-files/archive/master.zip, we don't need the entire history.

pyxdf/test/test_data.py
```diff
@@ -0,0 +1,68 @@
...
+        t = np.array([5., 5.1, 5.2, 5.3, 5.4, 5.5, 5.6, 5.7, 5.8])
+        np.testing.assert_array_equal(streams[0]["time_series"], s)
+        np.testing.assert_array_almost_equal(streams[0]["time_stamps"], t)
+
+        assert streams[1]["info"]["name"][0] == "SendDataString"
+        assert streams[1]["info"]["type"][0] == "StringMarker"
+        assert streams[1]["info"]["channel_count"][0] == "1"
+        assert streams[1]["info"]["nominal_srate"][0] == "10"
+        assert streams[1]["info"]["channel_format"][0] == "string"
+        assert streams[1]["info"]["stream_id"] == 46202862
```
tstenner: This should the literal `0x02C0FFEE` so it matches the writer test.

pyxdf/test/test_data.py
```diff
@@ -0,0 +1,68 @@
...
+extensions = ["*.xdf", "*.xdfz", "*.xdf.gz"]
+files = []
+for ext in extensions:
+    files.extend(path.glob(ext))
+files = [str(file) for file in files]
+
+
+@pytest.mark.parametrize("file", files)
+def test_load_file(file):
+    streams, header = load_xdf(file)
```
tstenner: When we have more files this could get unwieldy, but for now it's a good start.

cbrnr: What exactly do you mean? This works with any number of test files because it's simply loading them all. If you meant the specific tests, yes, this will need to be refactored into separate tests sooner or later.

pyxdf/test/test_data.py
cbrnr: OK - but where is this writer test?

azure-pipelines.yml
cbrnr: Done, changed to shallow clone.

pyxdf/test/test_data.py
tstenner: https://github.com/labstreaminglayer/App-LabRecorder/blob/master/src/test_xdf_writer.cpp

tstenner: Now that I think of it, there should be a link to it in the test file description

cbrnr: OK, done.

pyxdf/test/test_data.py
tstenner: It's 47 lines for a simple file and not feasible for longer files. For the first check if a file can be read correctly it's ok, but for other files it should test specific parts of the reader.

PR #50, opened 2019-10-18
Filesize detection doesn't work for compressed files

tstenner:
https://github.com/xdf-modules/xdf-python/blob/a6c5f84860c41de9d63fc3f66ac7cac0f7779bc7/pyxdf/pyxdf.py#L235-L245

For compressed files, reading the file is aborted as soon as  an invalid chunk size is read and the position in the *uncompressed* stream is after the *compressed* size. So, the new code checks if more than 1024 bytes are actually readable.


tstenner:
I have no idea where the 1024 bytes came from. I suspect that it wasn't considered too useful to find a boundary chunk if the file is too close to the end, but how the 1024 came from that I don't know.

cbrnr:
Can you add this explanation as a comment (it makes sense to not look for boundary chunks very close to the end of a file)?

tstenner:
It doesn't matter how much data is left, so I changed the check to read only one byte. If there's not enough left for a boundary chunk, `_scan_forward` will read until the end of the file and then quit. It's a tiny bit slower, but we've gotten rid of one magic number.

cbrnr:
And this happens only for corrupt files anyway, right?

tstenner:
Yes with a tiny bit of fineprint. For regular files it shouldn't happen.

pyxdf/pyxdf.py
```diff
@@ -242,11 +240,17 @@ def load_xdf(filename,
...
+                if f.read(1):
                     logger.warning('got zero-length chunk, scanning forward to '
                                    'next boundary chunk.')
-                    _scan_forward(f)
-                    continue
+                    # move the stream position one byte back
+                    f.seek(-1, 1)
+                    if _scan_forward(f):
+                        continue
+                    else:
```
cbrnr: I guess you could get rid of this `else: break` block because you have the same thing right in the next `else` block. I.e. I think you can remove the first `else` block completely (the one belonging to the `if _scan_forward(f)`), and also remove the second `else` and dedent the two subsequent lines.

tstenner: Right, I've changed it.

cbrnr:
Thanks @tstenner!

PR #51, opened 2019-10-18
Allow loading from already opened file objects

tstenner:
Use case: `f = pyxdf.pyxdf.open_xdf(io.BytesIO(b'XDF:'))`

More realistically, this allows us to construct test files in memory and load them for unit tests and end-users can use any compression scheme which has a python wrapper, e.g. `pyxdf.pyxdf.open_xdf(zstd.open('somefile.xdf.zstd'))`.

The first two commits are from #50 which is needed for obvious reasons.

### DIFF
--- a/pyxdf/pyxdf.py
+++ b/pyxdf/pyxdf.py
@@ -9,7 +9,7 @@
 
 This function is closely following the load_xdf reference implementation.
 """
-
+import io
 import struct
 import itertools
 import gzip
@@ -204,9 +204,6 @@ def load_xdf(filename,
         logger.setLevel(logging.DEBUG if verbose else logging.WARNING)
 
     logger.info('Importing XDF file %s...' % filename)
-    filename = Path(filename).resolve()  # absolute path following symlinks
-    if not filename.exists():
-        raise Exception('file %s does not exist.' % filename)
 
     # if select_streams is an int or a list of int, load only streams
     # associated with the corresponding stream IDs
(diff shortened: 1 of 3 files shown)


pyxdf/pyxdf.py
```diff
@@ -368,15 +366,28 @@ def load_xdf(filename,
...
+        if isinstance(file, io.TextIOBase):
+            raise ValueError('file has to be opened in binary mode')
+        f = file
     else:
-        f = open(str(filename), 'rb')
+        filename = Path(file).resolve()  # absolute path following symlinks
+        if not filename.exists():
+            raise Exception('file %s does not exist.' % filename)
+
+        filename = Path(file)  # ensure convert to pathlib object
```
cbrnr: I'd do `filename = Path(file)` in line 380, and then check `if not filename.resolve().exists():` in order to avoid assigning to `filename` twice.

cbrnr:
Also, I'm going to merge #50 before this one, so I hope that the two commits won't cause a conflict (to be safe, you could remove these commits from this PR).

pyxdf/pyxdf.py
tstenner: Done.

tstenner:
They have the same SHA1 hashes, so git will take care of it unless I force-push a rebased branch.

tstenner:
Let me correct that: unless I force-push or the other branch is squashed before merging.

I've rebased the branch, so now it should work.

cbrnr:
Thanks @tstenner!

PR #62, opened 2020-07-17
Handle a StreamId that can not be parsed correctly

### LINKED ISSUE #61: error: unpack requires a buffer of 4 bytes
I get the following error when I try to read one of my `.xdf` files: 

```
~/lib/python3.7/site-packages/pyxdf/pyxdf.py in load_xdf(filename, select_streams, on_chunk, synchronize_clocks, handle_clock_resets, dejitter_timestamps, jitter_break_threshold_seconds, jitter_break_threshold_samples, clock_reset_threshold_seconds, clock_reset_threshold_stds, clock_reset_threshold_offset_seconds, clock_reset_threshold_offset_stds, winsor_threshold, verbose)
    258             log_str = ' Read tag: {} at {} bytes, length={}'.format(tag, f.tell(), chunklen)
    259             if tag in [2, 3, 4, 6]:
--> 260                 StreamId = struct.unpack('<I', f.read(4))[0]
    261                 log_str += ', StreamId={}'.format(StreamId)
    262             else:

error: unpack requires a buffer of 4 bytes
```

Although, I can easily read the same file using EEGLab in Matlab. Do you have any recommendations on how I can solve this?
Thanks.

agricolab:
Resolve #61

Issue #61 reports that a specific file could not be loaded, as `unpack requires a buffer of 4 bytes` in line 260.
Inspection of the provided file showed an misconfigured StreamId message after the footer. The message was an empty byte `b''` instead of 4-bytes, so it could not be unpacked properly to an little-endian unsigned int. This part is now wrapped within a try-catch.

I tested it with the offending file, and `load_xdf` now is able to load it fully.

### DIFF
--- a/pyxdf/pyxdf.py
+++ b/pyxdf/pyxdf.py
@@ -260,9 +260,7 @@ def load_xdf(filename,
             if tag in [2, 3, 4, 6]:
                 _streamid = f.read(4)
                 try:
-                    StreamId = struct.unpack("<I", _streamid)[0]
-                    log_str += ", StreamId={}".format(StreamId)
-                    logger.debug(log_str)
+                    StreamId = struct.unpack("<I", _streamid)[0]                    
                 except struct.error:
                     # we scan forward to next (hopefully) valid block in a bid
                     # to load as much of the file as possible
@@ -274,6 +272,10 @@ def load_xdf(filename,
                     logger.error(log_str)
                     _scan_forward(f)
                     continue
+                else:
+                    # to be executed if no exception was raised
+                    log_str += ", StreamId={}".format(StreamId)
+                    logger.debug(log_str)
 
             if StreamId is not None and select_streams is not None:
                 if StreamId not in select_streams:
--- a/pyxdf/pyxdf.py
+++ b/pyxdf/pyxdf.py
@@ -256,11 +256,16 @@ def load_xdf(filename,
             # read [Tag]
             tag = struct.unpack('<H', f.read(2))[0]
             log_str = ' Read tag: {} at {} bytes, length={}'.format(tag, f.tell(), chunklen)
+            StreamId = None
             if tag in [2, 3, 4, 6]:
-                StreamId = struct.unpack('<I', f.read(4))[0]
-                log_str += ', StreamId={}'.format(StreamId)
-            else:
-                StreamId = None
+                _streamid = f.read(4)
+                try:
+                    StreamId = struct.unpack("<I", _streamid)[0]
+                    log_str += ", StreamId={}".format(StreamId)
+                except struct.error:
+                    StreamId = "StreamId missing"
+                    log_str += ", StreamId was missing"
+
 
             logger.debug(log_str)
 
--- a/pyxdf/pyxdf.py
+++ b/pyxdf/pyxdf.py
@@ -262,12 +262,18 @@ def load_xdf(filename,
                 try:
                     StreamId = struct.unpack("<I", _streamid)[0]
                     log_str += ", StreamId={}".format(StreamId)
+                    logger.debug(log_str)
                 except struct.error:
-                    StreamId = "StreamId missing"
-                    log_str += ", StreamId was missing"
-
-
-            logger.debug(log_str)
+                    # we scan forward to next (hopefully) valid block in a bid
+                    # to load as much of the file as possible
+                    # If the StreamId could not be parsed correctly, it will be
+                    # None. We therefore also need to continue, because
+                    # otherwise we might end up in one the tag-specific branches
+                    # which expect a valid StreamId
+                    log_str += ", but StreamId is corrupt, scanning forward to next boundary chunk."
+                    logger.error(log_str)
+                    _scan_forward(f)
+                    continue
 
             if StreamId is not None and select_streams is not None:
                 if StreamId not in select_streams:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 ### Added
 - Add cython type hints (requires optional local compilation) ([#17](https://github.com/xdf-modules/xdf-python/pull/17) by [Tristan Stenner](https://github.com/tstenner)).
 
+### Fixed
+- Handle XDF files with corrupt chunk headers (missing stream IDs) more gracefully ([#62](https://github.com/xdf-modules/xdf-python/pull/62) by [Robert Guggenberger](https://github.com/agricolab))
+
 ### Changed
 - `load_xdf` now requires keyword-only arguments after the first two arguments ([#59](https://github.com/xdf-modules/xdf-python/pull/59) by [Christian Kothe](https://github.com/chkothe)).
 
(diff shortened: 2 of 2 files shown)


cbrnr:
Wait, what happens with data where the StreamID is missing? This could happen anywhere in the file and not just after the footer - do we want this?

agricolab:
This happens far from silent. 

When a StreamId can not be parsed,  it will be `None` and the logger will print the error `found likely XDF file corruption`. If this happens multiple times,  the logger will print this a lot. Addtionally, the logger will record StreamId missing as a debug message. We could upgrade this to an error message.

cbrnr:
And what happens to the data associated with a StreamId of `None`?

agricolab:
Good point. Let's assume the StreamId is corrupt and stays `None`, but some data is readable. As it stands right now,  it would try to append it to the `temp[None].time_series`, but fail because this was never initialized. This triggers a catch and prints the `found likely file corruption`. See
https://github.com/xdf-modules/xdf-python/blob/ed661d37ab2ad47c76ad110be4f2101612f4e571/pyxdf/pyxdf.py#L299

Larger than the issue at hand: After reading the code again,  i am a little bit more worried about what happens if the StreamId for the header chunk is corrupted. We might want to safeguard against  this. 

Would you prefer if the data would be  tried to be saved somehow?

cbrnr:
I think that we shouldn't try to modify our reader so that it can read some corrupted XDF files. If the original files are not valid XDF, it would be better to repair them (maybe with something as simple as a hex editor) - is this possible with the files from #61?

agricolab:
Thanks for the feedback. 

There are many reasons why a file can become corrupted, from disk rot to copy error to recording error.  When a file is corrupted, i think there are two options for a reader on how to proceed:
-  **fail** with a sensible and informative error message, 
- or try to **recover** as much data as feasible. 

I guess we both agree that this pull request should not change the behavior of `load_xdf`, and that the function should behave consistently. 

Considering that as of now, `load_xdf` already recovers from corrupt samples chunks (see [L301](https://github.com/xdf-modules/xdf-python/blob/master/pyxdf/pyxdf.py#L301)), i have the impression that the function already follows the **recover** approach. 

That's why i chose to follow the **recover** approach for corrupted SampleId chunks. My rationale was that I struggle to see why a corrupted byte a little bit earlier in the file should be handled differently. I can still see the validity of the option that `load_xdf` deliberately fails on corrupted chunks. In that case, i can rewrite and give a clear error message. But i feel that then there would be an inconsistency in how corrupted chunks are being handled, and we should decide on how to handle corrupted samples (and other) chunks, too. 

I guess that this issue requires a explicit decision on what approach `load_xdf` should follow.  I guess  a compromise would be to add an additional argument to `load_xdf`, to decide how corrupted chunks should be handled, but i feel that the function already has way to many arguments. The least we should offer is a more informative error message for corrupted SampleId chunks than of now.

cbrnr:
I certainly agree that if the function fails to load a file it should present an informative error message. However, I don't think that `load_xdf` already tries to recover (or should try to recover) corrupted XDF files - the line you've linked basically drops/ignores a corrupted chunk of data (and doesn't recover the data in that chunk). I do get your point though, this is different than throwing an error and not continuing with the data reading process.

If a chunk ID is missing we could basically do the same thing - drop/ignore the data and continue reading the file. Therefore, I'm fine with the change, but this doesn't *recover* the data (which I thought was the original intention).

agricolab:
Great, i think we have a common understanding. Sorry for the wording, i meant to recover the rest of the file. I assume we agree that recovering the actual data in the chunk when the StreamId could not be parsed is best left for a different tool than `load_xdf`, e.g. the HexEditor you recommended.

I revised what happens when the StreamId can not be parsed following this discussion. It now explicitly scans forward to the next chunk and tries to continue with the rest of the file. I also added an explanatory comment.

Using `python -c "import pyxdf; pyxdf.load_xdf('1_DM_13.xdf')"`, i now get the following error log:
`Read tag: 3 at 3670621 bytes, length=82388, but StreamId is corrupt, scanning forward to next boundary chunk`.

I hope this finds your and the other maintainers approval. Looking forward to your feedback.

cbrnr:
In theory, you could just skip the current chunk and start parsing the next one, because you have a valid `chunklen`. However, I agree that this might not be reliable when the StreamId is corrupted, so skipping forward to the next boundary chunk is a good idea. I have one specific comment, if this is addressed 👍  for merge.

pyxdf/pyxdf.py
```diff
@@ -256,13 +256,24 @@ def load_xdf(filename,
...
-                log_str += ', StreamId={}'.format(StreamId)
-            else:
-                StreamId = None
-
-            logger.debug(log_str)
+                _streamid = f.read(4)
+                try:
+                    StreamId = struct.unpack("<I", _streamid)[0]
+                    log_str += ", StreamId={}".format(StreamId)
+                    logger.debug(log_str)
```
cbrnr, line 265: These lines will not be reached if a `struct.error` is raised in the previous statement. Therefore, I'd move these lines to an `else` block.

agricolab: I guess an alternative using an else-block could be:

``` python
StreamId = None
if tag in [2, 3, 4, 6]:
    _streamid = f.read(4)
    try:
        StreamId = struct.unpack("<I", _streamid)[0]
    except struct.error:
        pass
        
    if StreamId is None:            
        # If the StreamId could not be parsed correctly, it will be
        # None. 
        # we scan forward to next (hopefully) valid block in a bid
        # to load as much of the file as possible. We also need to
        # continue, because otherwise we might end up in one
        # the tag-specific branches which expect a valid StreamId
        log_str += ", but StreamId is corrupt, scanning forward to next boundary chunk."
        logger.error(log_str)
        _scan_forward(f)
        continue
    else:
        # the StreamId was successfully unpacked
        log_str += ", StreamId={}".format(StreamId)
        logger.debug(log_str)


```
I didn't benchmark or inspect the bytecode, but this simply adds an if-clause and is therefore very likely the more expensive solution. I also have the impression it separates the logic. 

The implementation with try-catch uses directly the order of statements is for flow control to increase efficiency. Specifically,  lines 264-265  log at the debug level, but (as you said) only if the bytes were successfully unpacked. If instead the bytes could not be unpacked and a `struct.error` is raised, we jump to the `except` in line 266 and, log at the error level, scan_forward and continue in line 274-276

cbrnr: You can move these two lines to an `else` block belonging to the `try` block (`try`/`except`/`else`).

cbrnr:
```python
try:
    StreamId = struct.unpack("<I", _streamid)[0]
except struct.error:
    log_str += ", but StreamId is corrupt, scanning forward to next boundary chunk."
    logger.error(log_str)
    _scan_forward(f)
    continue
else:
    log_str += ", StreamId={}".format(StreamId)
    logger.debug(log_str)
```

agricolab: Awesome, use this so rarely, i haven't thought of it. Changed as requested.

cbrnr: In theory, this isn't really necessary here (you could remove the `else` and dedent the two lines), but I think it makes the structure/intention clear.

cbrnr:
@cboulay or @tstenner do you have time to take a look?

cbrnr:
Thanks @agricolab!

agricolab:
Thanks for taking the time to review.

PR #63, opened 2020-07-23
Fix PEP8

cbrnr:
I've fixed all PEP8 complaints, hopefully without impacting anything else.

### DIFF
--- a/pyxdf/__init__.py
+++ b/pyxdf/__init__.py
@@ -10,3 +10,4 @@ except DistributionNotFound:  # package is not installed
     __version__ = None
 from .pyxdf import load_xdf, resolve_streams, match_streaminfos
 
+__all__ = [load_xdf, resolve_streams, match_streaminfos]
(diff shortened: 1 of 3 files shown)


agricolab:
Is there a consensus on which formatter to use for xdf-python, e.g. `autopep8`, `yapf` or `black` or is the formatting kept in shape manually?

cbrnr:
I usually do it manually, but I'd be fine with using any of the tools you've mentioned (`black` seems to be the most popular choice at the moment).

agricolab:
I read through it and could find no fault. Additionally, pytest runs fine on my machine, too.
Do you want to run an autoformatter first? I personally prefer black.

cbrnr:
Alright, the code is now black auto-formatted (with `-l 79` which is PEP8-compliant, although I'm starting to think that something like 92 lines like in Julia might be a more sensible character limit, but let's not discuss this here and now). @agricolab if you're happy feel free to merge.

PR #66, opened 2020-08-07
Prepare for v1.16.3 release

cbrnr:
I think it's a good time for a new release. Any objections?


cbrnr:
@cboulay @agricolab @tstenner I'd like to cut this release today if possible, so unless I don't hear from you within the next two hours I will go ahead and release.

cboulay:
#67 needs to be fixed because in its current state load_xdf can return incorrect timestamps for non-integer sampling rates.

cbrnr:
I'll submit a PR with a fix and then go ahead and make the release.

PR #73, opened 2020-12-29
Fix robust_fit for large timestamps by subtracting offset

cboulay:
Fixes issue reported [here](https://github.com/labstreaminglayer/App-LabRecorder/issues/44) and by user in Slack.

Briefly, very large timestamp values (especially on systems using seconds since 1970) lead to instabilities due to insufficient numerical precision. Here we subtract a constant from the timestamps before calculating the fit between timestamps and offsets. The offset is added back before returning.

There was also a little extra to be done to keep winsor_threshold functioning properly.

Sister PR in Matlab: https://github.com/xdf-modules/xdf-Matlab/pull/9

### DIFF
--- a/pyxdf/pyxdf.py
+++ b/pyxdf/pyxdf.py
@@ -611,11 +611,12 @@ def _clock_sync(
                 if range_i[0] != range_i[1]:
                     start, stop = range_i[0], range_i[1] + 1
                     e = np.ones((stop - start,))
-                    X = np.column_stack([e, clock_times[start:stop]])
-                    X /= winsor_threshold
-                    y = clock_values[start:stop] / winsor_threshold
+                    X = np.column_stack([e, np.array(clock_times[start:stop]) / winsor_threshold])
+                    y = np.array(clock_values[start:stop]) / winsor_threshold
                     # noinspection PyTypeChecker
-                    coef.append(_robust_fit(X, y))
+                    _coefs = _robust_fit(X, y)
+                    _coefs[0] *= winsor_threshold
+                    coef.append(_coefs)
                 else:
                     coef.append((clock_values[range_i[0]], 0))
 
@@ -702,6 +703,8 @@ def _robust_fit(A, y, rho=1, iters=1000):
     http://www.stanford.edu/~boyd/papers/distr_opt_stat_learning_admm.html
 
     """
+    offset = np.min(A[:, 1])
+    A[:, 1] = A[:, 1] - offset  # Don't use -= operator here!
     Aty = np.dot(A.T, y)
     L = np.linalg.cholesky(np.dot(A.T, A))
     U = L.T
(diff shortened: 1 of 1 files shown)


pyxdf/pyxdf.py
```diff
@@ -702,6 +703,8 @@ def _robust_fit(A, y, rho=1, iters=1000):
     http://www.stanford.edu/~boyd/papers/distr_opt_stat_learning_admm.html
 
     """
+    offset = np.min(A[:, 1])
+    A[:, 1] = A[:, 1] - offset  # Don't use -= operator here!
```
tstenner, line 707: Why not `-=`?

```
In [1]: import numpy as np

In [2]: A=np.random.rand(5,2)

In [3]: A
Out[3]: 
array([[0.894628  , 0.03031772],
       [0.15520362, 0.78182106],
       [0.02175627, 0.70150644],
       [0.42385584, 0.4072398 ],
       [0.77578102, 0.10038374]])

In [4]: A[:,1]-=np.min(A[:,1])

In [5]: A
Out[5]: 
array([[0.894628  , 0.        ],
       [0.15520362, 0.75150334],
       [0.02175627, 0.67118871],
       [0.42385584, 0.37692208],
       [0.77578102, 0.07006602]])
```

pyxdf/pyxdf.py
```diff
@@ -611,11 +611,12 @@ def _clock_sync(
                 if range_i[0] != range_i[1]:
                     start, stop = range_i[0], range_i[1] + 1
                     e = np.ones((stop - start,))
-                    X = np.column_stack([e, clock_times[start:stop]])
-                    X /= winsor_threshold
-                    y = clock_values[start:stop] / winsor_threshold
+                    X = np.column_stack([e, np.array(clock_times[start:stop]) / winsor_threshold])
```
tstenner, line 614: The old code divides e by winsor_threshold, the new code doesn't. Is this intended?

pyxdf/pyxdf.py
cbrnr: Probably because it operates on a slice? https://stackoverflow.com/a/9047211/1112283
In any case, the comment should explain why `-=` should not be used!

cboulay: I was trying to not assume that the second column of A was not shared somewhere else so I didn't want to mutate data in-place. But, this is a bit silly because I am still making many other assumptions that are much more likely to be invalidated than the above assumption. For example, I'm assuming that A will never be reused so it's ok to mutate its second column.

The better solution is to first copy A before modifying it. This is now updated.

pyxdf/pyxdf.py
cboulay: Yes, this is intended. It's also why the intercept is multiplied by winsor_threshold after the function returns. I had an embarrassingly hard time trying to get the offset adjustment within robust_fit to work when the ones column was scaled. I settled on leaving it as ones, making the offset adjustment rather straightforward.

cboulay:
Before review of #74 , I'd like to this to be reviewed and merged. Are there any outstanding concerns?

tstenner:
> Are there any outstanding concerns?

Yes, it hasn't been merged yet :-)

cbrnr:
Thanks @cboulay!

PR #82, opened 2021-12-10
Fix #81

### LINKED ISSUE #81: Error when dejitter_timestamps=False on stream without samples
When calling load_xdf(... dejitter_timestamps=False) on a file containing a stream without any samples, an IndexError is thrown in line pyxdf.py:384

`duration = stream.time_stamps[-1] - stream.time_stamps[0]
IndexError: index -1 is out of bounds for axis 0 with size 0`

This is because the first and last element of this empty array are accessed. For that matter, if the stream had only a single sample, duration would be 0 and the next line `stream.effective_srate = len(stream.time_stamps) / duration` would cause a division by 0 error.

This could be fixed by adding a condition
```
# Default effective sampling rate to 0 if stream does not contain any samples
if len(stream.time_stamps) <= 1:
    stream.effective_srate = 0
    break
```
I couldn't manage to open a new pull request, but please let me know if you need further information. I'm using pyxdf==1.16.3 but the error seems to be still in #5384490

agricolab:


pyxdf/pyxdf.py
```diff
@@ -611,7 +614,13 @@ def _clock_sync(
                 if range_i[0] != range_i[1]:
                     start, stop = range_i[0], range_i[1] + 1
                     e = np.ones((stop - start,))
-                    X = np.column_stack([e, np.array(clock_times[start:stop]) / winsor_threshold])
+                    X = np.column_stack(
+                        [
+                            e,
```
cbrnr, line 619: If you are reformatting anyway, you could remove `e` and directly use `np.ones((stop - start,))` here.

agricolab: sorry, that was my autoformatter.

agricolab: fixed it anyways

PR #87, opened 2023-01-04
Remove deprecated np.object type

cbrnr:
Replacing `np.object` with `object` should do the trick. This is required for NumPy >= 1.24.


cbrnr:
We should also make a new release after this is merged.

cbrnr:
This change should be fine, so I went ahead and merged it. I will also make a new release, because people will start upgrading to NumPy 1.24.

PR #95, opened 2023-12-13
Add a playback-over-LSL example script.

cboulay:
This moves the previous `example.py` to `print_metadata.py` in a new `examples` module.

Additionally, in the new module, I added playback_lsl.py which will take an arbitrary XDF file and play it back over LSL. This isn't meant to be a perfect reproduction of the original streams. e.g., the stream metadata only copies over some known keys. And I keep post-processing enabled so we eliminate some jitters. It's also inefficient to run everything in a single thread. The goal here is just to have something very convenient.


cboulay:
I've been using this for a while without issue.

PR #99, opened 2024-02-10
avoid crash on corrupt footer

mcvain:
equivalent of https://github.com/xdf-modules/xdf-Matlab/pull/19 but for pyxdf
closes https://github.com/labstreaminglayer/App-LabRecorder/issues/107

### DIFF
--- a/pyxdf/pyxdf.py
+++ b/pyxdf/pyxdf.py
@@ -13,7 +13,7 @@ import io
 import struct
 import itertools
 import gzip
-from xml.etree.ElementTree import fromstring
+from xml.etree.ElementTree import fromstring, ParseError
 from collections import OrderedDict, defaultdict
 import logging
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


pyxdf/pyxdf.py
```diff
@@ -328,7 +328,16 @@ def load_xdf(
...
+                try:
+                    streams[StreamId]["footer"] = _xml2dict(fromstring(xml_string))
+                except Exception as e:
+                    if type(e).__name__ == 'ParseError':
+                        logger.error(
+                            "found likely XDF file corruption (%s), "
+                            "ignoring corrupted XML element in footer." % e
+                        )
+                    else:
+                        raise
```
cbrnr, line 340: Can you simplify this by catching only a `ParseError` exception in the first place? No need for the `if`/`else` clause then. Thanks!

mcvain: Okay. Thank you!

cbrnr: One last nitpick, would you mind using an f-string?

mcvain: I am happy to, I just thought I should be consistent with the other error messages. Let me know if it looks good!

cbrnr:
Thanks a lot @mcvain!

PR #100, opened 2024-04-12
Bump to Python 3.9

cbrnr:
Although we already use `importlib.metadata` (the recommended approach), we only caught `ImportError` when trying to `import pkg_resources`. However, if `setuptools` is installed, `pkg_resources` can still be imported, but raise a `DeprecationWarning`. This PR bumps the minimum required Python version to 3.9, which ships with `importlib.metadata`.


cboulay:
Looks good but I think `pyxdf` itself needs a version bump too. I guess that comes from the tag-and-release process.

cboulay (approved):

cbrnr:
Yes, I think we should make a new release after this PR is merged (we need a changelog entry then, I'll add one).

PR #104, opened 2024-04-18
Do not package the tests in the bdist/wheel

musicinmybrain:
It seems like this was the intent from the beginning, but the pattern `tests*` in the `excludes` for `find_package` does not match anything. This PR adjusts that pattern to `pyxdf.test*` so that the tests are not unnecessarily included with wheel/binary distributions.

The other two entries, `contrib` and `docs`, could also be removed, since they don’t match anything either.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -52,7 +52,7 @@ setup(
         "biosignals",
         "stream",
     ],
-    packages=find_packages(exclude=["contrib", "docs", "pyxdf.test*"]),
+    packages=find_packages(exclude=["pyxdf.test*"]),
     install_requires=["numpy"],
     extras_require={},
     package_data={},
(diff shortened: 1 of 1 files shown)


cbrnr:
I'm not sure that we want to exclude tests from the distribution package. I know that opinions differ, but I tend to think that having tests with an installed package can be useful.

Having said that, some tests provided with pyxdf require testing data from a different repository, so they would not work out of the box. Therefore, I'm fine with removing the tests (for now). Can you also remove the other two unused entries?

Eventually, we should switch to `pyproject.toml` BTW.

musicinmybrain:
Thanks for the feedback. That’s a reasonable (and reasonably nuanced) perspective.

I’ve removed the unused excludes in a follow-up commit.

PR #108, opened 2024-07-17
Add release GitHub Action

### LINKED ISSUE #103: Replace Azure Pipelines with GitHub Actions
Our Azure Pipeline workflows have been dysfunctional for some time now. This means that neither tests nor publishing on PyPI works. We should replace this with suitable GitHub Actions and restore automatic publishing whenever a new release is created.

cbrnr:
Fixes #103. This needs to be tested, since I've basically just copied the yml from MNELAB. In particular, we need to make sure that we've added the correct access token to upload to PyPI.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,47 @@
+name: Release
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  style:
+    name: Check style
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - uses: chartboost/ruff-action@v1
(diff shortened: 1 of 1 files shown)


cbrnr:
I think I've set up the required trusted publisher info in pyxdf's PyPI settings, so *maybe* we're good to go?

cboulay:
I don't mind if you stumble through a few minor version bumps while setting up the release. Go ahead and merge and release and let's see!

If you don't want to do that then you can try 
```
with:
    repository-url: https://test.pypi.org/legacy/
```
(After creating credentials on test.pypi.org for this package)

Issue #116, opened 2024-10-08
Information about segments is  not exposed to the user


agricolab:
When applying dejitter_timestamps (which is the default), it can happen that the streams are segmented and segments are dejittered piece-wise. The start and stop of the segments is not exposed to the user, at least not as far as i can tell.

PR #117, opened 2024-10-08
Expose information about segments in info and "print_metadata"

agricolab:
Addresses #116

### DIFF
--- a/pyxdf/examples/print_metadata.py
+++ b/pyxdf/examples/print_metadata.py
@@ -16,7 +16,7 @@ def main(fname: str):
 
     print("Found {} streams:".format(len(streams)))
     for ix, stream in enumerate(streams):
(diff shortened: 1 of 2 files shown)


cbrnr:
LGTM! Would it make sense to adapt the effective sampling frequency though? Currently, segments are ignored, which can lead to deviations if there are long(ish) gaps in the data. It would probably make more sense to compute an effective sampling frequency per segment and then average across segments?

agricolab:
If one dejitters, calculation of the effective srate as weighted average across segments is the default anyways.

agricolab:
So if it's okay, passes tests and has no conflicts, I'd merge this afternoon. Any opposing argument?

cbrnr:
Thanks @agricolab, I went ahead and merged!

cbrnr:
(I pushed a quick changelog entry after merging.)

PR #118, opened 2024-10-21
Rename examples to cli

cbrnr:
I think `pyxdf.examples` should be named `pyxdf.cli`, as this is what it really contains.

In addition, I've enforced a line length of 88 (default of Black/Ruff).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,10 +8,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
-    - uses: chartboost/ruff-action@v1
-    - uses: chartboost/ruff-action@v1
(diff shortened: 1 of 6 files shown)


cboulay (approved):
Looks good, thanks!

sappelhoff:
IMHO https://github.com/xdf-modules/xdf-tagger could also live here, rather than in its own repo. It is very similar in scope to the other "examples".

cbrnr:
@sappelhoff yes, I think so too! Would you be willing to submit a separate PR? I'd like to merge this one before adding new things.

sappelhoff:
> Would you be willing to submit a separate PR? I'd like to merge this one before adding new things.

sure, I could do that. However, I think this should go hand in hand with updating the README and then archiving said repo, to avoid duplicates

PR #119, opened 2024-10-30
FIX: Not looping leading to TypeError

sappelhoff:
A flag like `--loop` in this case is typically used as a "true-if-supplied" flag. However, it was instead (~~confusingly~~ counter-intuitively) used as a "false-if-supplied" flag.

This is now fixed. `python -m pyxdf.examples.playback_lsl.py --loop` will loop, whereas not supplying `--loop` will not loop.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,11 @@
 ## [UNRELEASED] - YYYY-MM-DD
 ### Fixed
 - Expose detected segment (used in dejittering) as `stream["info"]["segments"]` ([#117](https://github.com/xdf-modules/pyxdf/pull/117) by [Robert Guggenberger](https://github.com/agricolab))
-- Fix logic of `--loop` argument in `pyxdf.examples.playback_lsl.py`: Supplying `--loop` will loop, whereas ommitting `--loop` will NOT loop ([#119](https://github.com/xdf-modules/pyxdf/pull/119) by [Stefan Appelhoff](https://github.com/sappelhoff))
+- A non-looping playback of an XDF file will no longer lead to a `TypeError` ([#119](https://github.com/xdf-modules/pyxdf/pull/119) by [Stefan Appelhoff](https://github.com/sappelhoff))
+
+### Changed
+- Reverse logic of `--loop` argument in `pyxdf.examples.playback_lsl.py` to be more in line with standard practice: Supplying `--loop` will loop, whereas ommitting `--loop` will NOT loop ([#119](https://github.com/xdf-modules/pyxdf/pull/119) by [Stefan Appelhoff](https://github.com/sappelhoff))
+
 
 
 ### Changed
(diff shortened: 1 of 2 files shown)


sappelhoff:
Actually, there is more wrong. Not looping is not possible currently:

```
  File "C:\Users\stefan\AppData\Local\miniforge3\envs\naf\Lib\site-packages\pyxdf\examples\playback_lsl.py", line 229, in <module>
    main(
  File "C:\Users\stefan\AppData\Local\miniforge3\envs\naf\Lib\site-packages\pyxdf\examples\playback_lsl.py", line 192, in main
    timer.update()
  File "C:\Users\stefan\AppData\Local\miniforge3\envs\naf\Lib\site-packages\pyxdf\examples\playback_lsl.py", line 103, in update
    overrun = self._n_loop * self._boundary
              ~~~~~~~~~~~~~^~~~~~~~~~~~~~~~
TypeError: unsupported operand type(s) for *: 'int' and 'NoneType'
```

sappelhoff:
@cbrnr I'd be happy to receive a review on this!

cboulay:
playback_lsl is mine, but I'm happy to see others use it and improve it.

The previous default behaviour was to loop, and now the default behaviour is to not loop. I prefer if the default is to loop, but then how do we fix the argument name in argparse? Do we only offer "--no-loop"?

I'm not great with `argparse`. I tend to use `typer` more but I don't think we need to add that as a dependency unless we really plan to expand the CLI offerings. Anyway, in `typer`, bool args get exposed both in the affirmative and negative. e.g., this would give use `--loop` and `--no-loop`; both would be exposed regardless of what we set the default to.

cboulay:
I can also be convinced that not-looping is the better default behaviour.

sappelhoff:
Hey @cboulay -- thanks for programming and sharing this script, it has been very helpful for me!

> Do we only offer "--no-loop"?

yes, that would be a sensible alternative solution.

> I can also be convinced that not-looping is the better default behaviour.

I think from a user experience perspective it's more natural to expect a playback to NOT loop by default (think of music that you play).

Please let me know if you want me to change the default back to looping, and adjust the flag to be `--no-loop`.

Other than that, it would be ready for merge from my side :)

cbrnr:
Just my two cents (and you said you can be convinced), I'd also think that _not_ looping is a better default. But I leave this up to you, so feel free to merge whenever you're happy @cboulay!

cbrnr:
@sappelhoff can you rebase please? Afterwards I think we can merge.

sappelhoff:
@cbrnr sure -- done!

CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 ## [UNRELEASED] - YYYY-MM-DD
 ### Fixed
 - Expose detected segment (used in dejittering) as `stream["info"]["segments"]` ([#117](https://github.com/xdf-modules/pyxdf/pull/117) by [Robert Guggenberger](https://github.com/agricolab))
+- A non-looping playback of an XDF file will no longer lead to a `TypeError` ([#119](https://github.com/xdf-modules/pyxdf/pull/119) by [Stefan Appelhoff](https://github.com/sappelhoff))
+
+### Changed
```
cbrnr, line 6: The rebase already added a Changed section.

cbrnr:
Thanks @sappelhoff!

PR #121, opened 2024-11-22
Fix/expose segments

jamieforth:
This addresses a couple of small issues with segment metadata.

When `dejitter_timestamps=False` reported segments now have the correct sample stop. 

When `dejitter_timestamps=True`
- non-empty streams no longer contain duplicate segment tuples
- empty stream now return segments as an empty array, rather than [0, -1]

### DIFF
--- a/pyxdf/pyxdf.py
+++ b/pyxdf/pyxdf.py
@@ -348,7 +348,7 @@ def load_xdf(
             if stream.fmt == "string":
                 stream.time_series = []
             else:
-                stream.time_series = np.zeros((stream.nchns, 0))
+                stream.time_series = np.zeros((0, stream.nchns))
 
     # perform (fault-tolerant) clock synchronization if requested
     if synchronize_clocks:
--- a/pyxdf/pyxdf.py
+++ b/pyxdf/pyxdf.py
@@ -364,10 +364,6 @@ def load_xdf(
         )
 
     # perform jitter removal if requested
-    for tmp in temp.values():
-        # initialize segment list in case jitter_removal was not selected
-        tmp.segments = [(0, len(stream.time_series) - 1)]  # inclusive
-
     if dejitter_timestamps:
         logger.info("  performing jitter removal...")
         temp = _jitter_removal(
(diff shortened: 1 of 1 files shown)


cbrnr:
@agricolab do you have time to look?

pyxdf/pyxdf.py
```diff
@@ -629,6 +629,7 @@ def _jitter_removal(streams, threshold_seconds=1, threshold_samples=500):
     for stream_id, stream in streams.items():
         stream.effective_srate = 0  # will be recalculated if possible
         nsamples = len(stream.time_stamps)
+        stream.segments = []
```
agricolab, line 632: 👌🏼 catches empty streams and returns [] instead of [0,-1] (when `_jitter_removal` was called, i.e. `dejitter_timestamps=True`

pyxdf/pyxdf.py
```diff
@@ -348,7 +348,7 @@ def load_xdf(
             if stream.fmt == "string":
                 stream.time_series = []
             else:
-                stream.time_series = np.zeros((stream.nchns, 0))
+                stream.time_series = np.zeros((0, stream.nchns))
```
agricolab, line 351: 👌🏼  good catch. we probably need a test for empty streams eventually to catch such issues

agricolab:
Did a manual test and it passes now unduplicated segments

```
import pyxdf; 
s, h = pyxdf.load_xdf("minimal.xdf", jitter_break_threshold_seconds=0.001, jitter_break_threshold_samples=1);
assert s[0]["info"]["segments"] == [(0, 0), (1, 3), (4, 8)]
```

agricolab:
Thanks @jamieforth !

cbrnr:
Thanks @jamieforth and @agricolab! Does the test you mention pass with this change, but fail previously? If so, could you add it to `pyxdf/test/test_data.py`?

agricolab:
Yes, could be a good regression test. I'd also like if we'd had a minimal.xdf (or empty.xdf) with an empty stream.

PR #126, opened 2024-12-05
Migrate to `uv` builds

cboulay:
Note: Using "trusted publisher" between pypi <--> GitHub, the `uv build` and `uv publish` steps are quite simple and don't require tokens or an intermediate artifact.

__version__.py file written upon build.


.github/workflows/release.yml
```diff
@@ -17,31 +17,15 @@ jobs:
 
   build:
     needs: style
-    name: Build package
+    name: Build and publish package
     runs-on: ubuntu-22.04
```
cbrnr, line 21:
```suggestion
    runs-on: ubuntu-24.04
```

.github/workflows/test.yml
```diff
@@ -1,6 +1,13 @@
 name: Test
 
-on: [push, pull_request]
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches:
+      - main
+      - dev
```
cbrnr, line 9: I didn't even know that we're using the `dev` branch. I'd prefer merging everything in `main`, but if you prefer a different model, we should describe it somewhere.

.github/workflows/test.yml
```diff
@@ -17,22 +24,20 @@ jobs:
     needs: style
     strategy:
       matrix:
-        python-version: ['3.9', '3.10', '3.11', '3.12']
+        python-version: ["3.9", "3.10", "3.11", "3.12"]
```
cbrnr, line 27: What about 3.13?

pyproject.toml
```diff
@@ -0,0 +1,76 @@
...
+  "Operating System :: Microsoft :: Windows",
+  "Operating System :: POSIX :: Linux",
+  "Operating System :: MacOS",
+
+  # Specify the Python versions you support here.
+  "Programming Language :: Python :: 3",
+  "Programming Language :: Python :: 3.9",
+  "Programming Language :: Python :: 3.10",
+  "Programming Language :: Python :: 3.11",
+  "Programming Language :: Python :: 3.12",
```
cbrnr, line 46: Please remove the comments.

pyproject.toml
```diff
@@ -0,0 +1,76 @@
...
+Issues = "https://github.com/xdf-modules/pyxdf/issues"
+Changelog = "https://github.com/xdf-modules/pyxdf/blob/main/CHANGELOG.md"
+
+[project.optional-dependencies]
+test = [
+    "pytest>=8.3.4",
+]
+dev = [
+    "ruff>=0.8.2",
+]
```
cbrnr, line 63: I'd only create a single `dev` extra here. Or, maybe even better, you could use the `dev` dependency group (https://docs.astral.sh/uv/concepts/projects/dependencies/#dependency-groups).

test/__init__.py
cbrnr, line 1: This is not needed.

src/pyxdf/__init__.py
```diff
@@ -0,0 +1,4 @@
+from .__version__ import __version__ as __version__
+from .pyxdf import load_xdf, match_streaminfos, resolve_streams
+
+__all__ = [load_xdf, resolve_streams, match_streaminfos]
```
cbrnr, line 4: `__all__` should be a list of strings.

cbrnr:
After this PR is merged, I'd like to cut a new release (which is a nice test if the new `uv publish` works). I think we should bump to 1.17.0 this time.

.github/workflows/test.yml
cboulay: I think adding 3.13 can be a separate issue / PR. Maybe there are some numpy version issues we'll have to deal with, and that effort is orthogonal to this effort.

pyproject.toml
cboulay: I just realized that you meant you'd like both pytest and ruff in a single group.

Sometimes the dev group can get pretty big and you don't want the test runner to import and install all those dependencies. But we aren't there yet. I can make it a single group...

cboulay:
I'm in transit today. If there are any finishing touches you want to do then please go ahead without me.

cbrnr:
This looks good to me except for one point: I'm not entirely convinced that we should include `uv.lock` under version control. I know that the general recommendation is to include it, but I view pyxdf as a library (as opposed to an application), in which case I'd rather not limit a user's environment, since they will use pyxdf in combination with many other packages that we cannot (and don't want to) control.

cboulay:
OK ready to go. You're welcome to squash if you want when merging.

cbrnr:
Thanks @cboulay!

PR #127, opened 2024-12-07
Segment at negative timestamp breaks

jamieforth:
When dejittering, also take into account negative timestamps when
determining segment boundaries.

### DIFF
--- a/src/pyxdf/pyxdf.py
+++ b/src/pyxdf/pyxdf.py
@@ -633,7 +633,7 @@ def _jitter_removal(streams, threshold_seconds=1, threshold_samples=500):
         if nsamples > 0 and stream.srate > 0:
             # Identify breaks in the time_stamps
             diffs = np.diff(stream.time_stamps)
-            b_breaks = diffs > np.max(
+            b_breaks = np.abs(diffs) > np.max(
                 (threshold_seconds, threshold_samples * stream.tdiff)
             )
             # find indices (+ 1 to compensate for lost sample in np.diff)
--- a/src/pyxdf/pyxdf.py
+++ b/src/pyxdf/pyxdf.py
@@ -625,19 +625,26 @@ def _clock_sync(
     return streams
 
 
-def _jitter_removal(streams, threshold_seconds=1, threshold_samples=500):
+def _detect_breaks(stream, threshold_seconds=1.0, threshold_samples=500):
+    """Detect breaks in the time_stamps of a stream."""
+    # Identify breaks in the time_stamps
+    diffs = np.diff(stream.time_stamps)
+    b_breaks = np.abs(diffs) > np.max(
+        (threshold_seconds, threshold_samples * stream.tdiff)
+    )
+    # find indices (+ 1 to compensate for lost sample in np.diff)
+    break_inds = np.where(b_breaks)[0] + 1
+    return break_inds
+
+
+def _jitter_removal(streams, threshold_seconds=1.0, threshold_samples=500):
     for stream_id, stream in streams.items():
         stream.effective_srate = 0  # will be recalculated if possible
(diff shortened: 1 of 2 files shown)


agricolab:
Interesting catch. If i understand this correctly, this would catch streams where timestamp progression is negative (i.e. sampling rate was negative), not necessarily streams with negative timestamps, right?

` np.diff([-4.-2,-1,0,1,2])`  -> `array([2, 1, 1, 1, 1])`
but 
`np.diff([2,1,0,-1,-2,-4])` -> `array([-1, -1, -1, -1, -2])`

Is that would you wanted?

agricolab:
Consider also a possible refactoring and adding tests like  in

https://github.com/agricolab/xdf-Python/tree/pr/fix/expose-segments

agricolab (approved):
I did a refactoring and added tests (see 
https://github.com/agricolab/xdf-Python/tree/pr/fix/expose-segments)

to see whether it would cause any breaking changes. Passes the new tests and this implementation additionally works for negative time and glitches, now too

jamieforth:
> If i understand this correctly, this would catch streams where timestamp progression is negative (i.e. sampling rate was negative), not necessarily streams with negative timestamps, right?

Exactly, negative time *intervals* would have been a better way to put it, sorry.

> unless we take the abs as you suggest.
> 
> Is that would you wanted? Or should we handle glitches, but throw an error for negative time?

Thank you for the tests, this really helped clarify things.

I don't see a problem with handling negative time or negative intervals per se, then it's up to the user to decide what to do with them. However, I now think there might be a better way to do it.

So I retract the `abs` suggestion in favour of explicitly handling non-positive time interval. Then the original logic for positive time intervals remains the same. 

The strategy I've gone for here is to always segment at a non-positive time interval – regardless of the thresholds (i.e. segments are by definition monotonic).

I can't really imagine a use case for non-monotonic segments. But if you think it's a too risky breaking change, perhaps something like `_detect_breaks(monotonic=False)` could preserve the existing default behaviour (in conjunction with `abs`).

jamieforth:
Another rethink... An underling problem to all the above is data being recorded out-of-order (which I don't think itself is necessarily a problem, just something more likely to happen with busy/WiFi networks etc.).

So ensuring the time-series data is sorted prior to synchronisation and dejitter avoids the problem!

agricolab:
Thanks for the effort and the additonal work. 

I agree that non-monoticity is clearly indicative of a segmentation. Now practically e.g `[0, 1, 2, 5, 6, 3, 4]` would currently only catch the second  segment `[5,6]` but not the last `[3,4]`. 

Sorting this prior to segmentation could be an option, but i would not make it the default for the following reasons. 

First, unsorted (i.e. non-monotic) timestamps  would in principle caught as  multiple segments, when we catch negative tdiffs in `_detect_breaks`. The difference between sorting before versus after segmentation is that as segments are  dejittered segment-wise, they are dejittered differently. Therefore, sorted first and sorted last will share the identical timeseries, but the respective timestamps will be different. Might or might not be what the user wants - but that means we need to be able to set the option.

Second, pyxdf policy is to be hands-off the timeseries. We discussed this a lot during the PR for syncing stamps of different streams. The idea is to be hands-off unless the user explicitly tells pyxdf to manipulate the timeseries (and not just the timestamps).  

In conclusion, my suggestion would be to catch `b_breaks = np.logical_or(diffs < 0, diff > np.max((threshold_seconds, threshold_samples * stream.tdiff)))` in `_detect_breaks` and  make sorting before segmentation an optional kwarg to load_xdf.

Additionally,  I think that non-monotonic timestamps might be linked to clock resets, so we should be aware that this is issue is also handled by `_clock_sync`.

jamieforth:
Closing this pull request, will create a new one from a separate branch.

PR #129, opened 2025-01-26
Ensure empty stream segments are initialised

jamieforth:
* Initialise empty segment list in StreamData

* Initialise empty stream segments when dejitter_timestamps=True

* Update load data tests and add test for a file with an empty streams

### DIFF
--- a/src/pyxdf/pyxdf.py
+++ b/src/pyxdf/pyxdf.py
@@ -378,7 +378,7 @@ def load_xdf(
         for stream in temp.values():
             if len(stream.time_stamps) > 1:
                 duration = stream.time_stamps[-1] - stream.time_stamps[0]
-                stream.effective_srate = len(stream.time_stamps) / duration
+                stream.effective_srate = (len(stream.time_stamps) - 1) / duration
             else:
                 stream.effective_srate = 0.0
             # initialize segment list in case jitter_removal was not selected
--- a/src/pyxdf/pyxdf.py
+++ b/src/pyxdf/pyxdf.py
@@ -633,7 +633,7 @@ def _jitter_removal(streams, threshold_seconds=1, threshold_samples=500):
         nsamples = len(stream.time_stamps)
         if nsamples > 0:
             if stream.srate == 0:
-                # Initialise default segment for irregular sampling rate streams.
+                # Initialise default segment for irregular sampling rate streams
                 stream.segments.append((0, nsamples - 1))  # inclusive
                 continue
 
--- a/src/pyxdf/pyxdf.py
+++ b/src/pyxdf/pyxdf.py
@@ -665,12 +665,11 @@ def _jitter_removal(streams, threshold_seconds=1, threshold_samples=500):
             # Recalculate effective_srate if possible
             counts = (seg_stops + 1) - seg_starts
             if np.any(counts):
-                # Calculate range segment duration (assuming last sample duration was
-                # exactly 1 * stream.tdiff)
+                # Calculate range segment duration
                 durations = (
-                    stream.time_stamps[seg_stops] + stream.tdiff
-                ) - stream.time_stamps[seg_starts]
-                stream.effective_srate = np.sum(counts) / np.sum(durations)
+                    stream.time_stamps[seg_stops] - stream.time_stamps[seg_starts]
+                )
+                stream.effective_srate = np.sum(counts - 1) / np.sum(durations)
 
             srate, effective_srate = stream.srate, stream.effective_srate
             if np.abs(srate - effective_srate) / srate > 0.1:
--- a/src/pyxdf/pyxdf.py
+++ b/src/pyxdf/pyxdf.py
@@ -664,7 +664,7 @@ def _jitter_removal(streams, threshold_seconds=1, threshold_samples=500):
 
             # Recalculate effective_srate if possible
             counts = (seg_stops + 1) - seg_starts
-            if np.any(counts):
+            if np.any(counts > 1):
                 # Calculate range segment duration
                 durations = (
                     stream.time_stamps[seg_stops] - stream.time_stamps[seg_starts]
--- a/src/pyxdf/pyxdf.py
+++ b/src/pyxdf/pyxdf.py
@@ -59,6 +59,9 @@ class StreamData:
         # nominal sampling interval, in seconds, for delta decompression
         self.tdiff = 1.0 / self.srate if self.srate > 0 else 0.0
         self.effective_srate = 0.0
+        # list of segments corresponding to detected time-stamp breaks
+        # (each a tuple of start_idx, end_idx)
+        self.segments = []
         # pre-calc some parsing parameters for efficiency
         if self.fmt != "string":
             self.dtype = np.dtype(fmts[self.fmt])
@@ -379,9 +382,8 @@ def load_xdf(
             else:
                 stream.effective_srate = 0.0
             # initialize segment list in case jitter_removal was not selected
-            stream.segments = []
             if len(stream.time_stamps) > 0:
-                stream.segments.append((0, len(stream.time_series) - 1))  # inclusive
+                stream.segments.append((0, len(stream.time_stamps) - 1))  # inclusive
 
     for k in streams.keys():
         stream = streams[k]
@@ -629,8 +631,12 @@ def _jitter_removal(streams, threshold_seconds=1, threshold_samples=500):
     for stream_id, stream in streams.items():
         stream.effective_srate = 0  # will be recalculated if possible
         nsamples = len(stream.time_stamps)
-        stream.segments = []
-        if nsamples > 0 and stream.srate > 0:
+        if nsamples > 0:
+            if stream.srate == 0:
+                # Initialise default segment for irregular sampling rate streams.
+                stream.segments.append((0, nsamples - 1))  # inclusive
+                continue
+
             # Identify breaks in the time_stamps
             diffs = np.diff(stream.time_stamps)
             b_breaks = diffs > np.max(
@@ -645,6 +651,7 @@ def _jitter_removal(streams, threshold_seconds=1, threshold_samples=500):
             seg_stops = np.hstack((break_inds - 1, nsamples - 1))  # inclusive
             for a, b in zip(seg_starts, seg_stops):
                 stream.segments.append((a, b))
+
             # Process each segment separately
             for start_ix, stop_ix in zip(seg_starts, seg_stops):
                 # Calculate time stamps assuming constant intervals within each segment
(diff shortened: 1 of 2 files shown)


cbrnr (commented):
Perfect, thank you! Just a small question/suggestion, you are now checking if a test file exists, and if not you're skipping the test, which is really useful. However, a `@pytest.mark.skipif` marker would probably be even more visible.

cbrnr:
Also, I saw that you made several PRs (great!), can we go one by one to make reviewing and merging easier? I suggest that we start with this one, and then you tell me which one to do next.

jamieforth:
> a `@pytest.mark.skipif` marker would probably be even more visible

The best I could come up with is to use a fixture.

`@pytest.mark.skipif` evaluates the condition at import time, which seems to end up requiring multiple hard-coded paths.

```
@pytest.mark.skipif(
    not Path("example-files/minimal.xdf").exists(),
    reason=f"File not available.",
)
def test_minimal_file():
    path = Path("example-files/minimal.xdf")
    ...
```

So the fixture seems like a better option to me?

jamieforth:
> Also, I saw that you made several PRs (great!), can we go one by one to make reviewing and merging easier? I suggest that we start with this one, and then you tell me which one to do next.

For sure. I've marked the others as draft for now. They are basically in sequential order - each PR builds on the previous. I've tried to break them up into separate issues to give you a heads up on where I was trying to get to, and yeah was thinking the same, take one at a time and rebase as we go!

cbrnr:
> The best I could come up with is to use a fixture.
> 
> `@pytest.mark.skipif` evaluates the condition at import time, which seems to end up requiring multiple hard-coded paths.
> 
> So the fixture seems like a better option to me?

Yes, I think it's a good option. However, to avoid having to duplicate file paths, you could define a path for each test file globally (since we seem to have exactly one test per file instead of a test using several different files). Something like

```python
path_minimal = Path("example-files/minimal.xdf")

@pytest.mark.skipif(not path_minimal.exists(), reason="File not found.")
def test_minimal_file():
    streams, header = load_xdf(path_minimal)
    ...
```

Actually, we already have a list of available testing files as `files`, so maybe we could use this list? Maybe we could explicitly add individual testing files (instead of wildcarding) and use a dictionary, which would make it more reusable?

```python
files = {
    "minimal": path / "minimal.xdf",
    "clock_resets": path / "data_with_clock_resets.xdf",
    "empty_streams": path / "data_with_empty_streams.xdf",
}
files = {key: path / value for key, value in files.items() if (path / value).exists()}

@pytest.mark.skipif("minimal" not in files, reason="File not found.")
def test_minimal_file():
    streams, header = load_xdf(files["minimal"])
    ...
```

I think I prefer the last option, WDYT?

While we're at it, I think the testing files should not start with `data_with_` and just be named `clock_resets.xdf` and `empty_streams.xdf`.

jamieforth:
I've added the dictionary of files idea it seems pretty clear, although I still don't like having to specify the key twice :)

I've dropped the `data_with_` file names.

I also found an off-by-one error in the calculation of effective sample-rate when skipping dejittering.

`clock_resets.xdf` still needs a test. Do you want to add one, or we could wait until we get to #131?

src/pyxdf/pyxdf.py
```diff
@@ -375,13 +378,12 @@ def load_xdf(
         for stream in temp.values():
             if len(stream.time_stamps) > 1:
                 duration = stream.time_stamps[-1] - stream.time_stamps[0]
-                stream.effective_srate = len(stream.time_stamps) / duration
+                stream.effective_srate = (len(stream.time_stamps) - 1) / duration
```
cbrnr, line 381: Are you sure it wasn't correct before? Below, the `- 1` is needed because it stores the start and stop indices of segments, but here we're calculating the duration. If you only have 1 sample, you will now always get an `effective_srate` of `0` (previously it was `1 / duration`, which I think is the correct value).

src/pyxdf/pyxdf.py
```diff
@@ -629,8 +631,12 @@ def _jitter_removal(streams, threshold_seconds=1, threshold_samples=500):
     for stream_id, stream in streams.items():
         stream.effective_srate = 0  # will be recalculated if possible
         nsamples = len(stream.time_stamps)
-        stream.segments = []
-        if nsamples > 0 and stream.srate > 0:
+        if nsamples > 0:
+            if stream.srate == 0:
+                # Initialise default segment for irregular sampling rate streams.
```
cbrnr, line 636:
```suggestion
                # Initialise default segment for irregular sampling rate streams
```

test/test_data.py
```diff
@@ -5,91 +5,388 @@
...
+    np.testing.assert_equal(streams[i]["clock_times"], clock_times)
+    np.testing.assert_equal(streams[i]["clock_values"], clock_values)
+
+    streams, header = load_xdf(
+        path,
+        synchronize_clocks=synchronize_clocks,
+        dejitter_timestamps=dejitter_timestamps,
+        jitter_break_threshold_seconds=0.001,
+        jitter_break_threshold_samples=1,
+    )
```
cbrnr, line 148: Is it necessary to load the file a second time? Can you adapt the test so that we only need to load it once at the very top?

test/test_data.py
```diff
@@ -5,91 +5,388 @@
...
+    if dejitter_timestamps:
         assert streams[0]["info"]["segments"] == [(0, 0), (1, 3), (4, 8)]
+    else:
+        assert streams[0]["info"]["segments"] == [(0, 8)]
+
+
+@pytest.mark.parametrize("dejitter_timestamps", [False, True])
+@pytest.mark.parametrize("synchronize_clocks", [False, True])
```
cbrnr, line 156: It seems like you do not use these two parameters in your test. Depending on if this was intentional, I suggest removing these parameterizations or adding tests for these settings.

cbrnr (commented):
I still have a couple of comments. Regarding a test for `clock_resets.xdf`, let's add it later (not in this PR). I think it's OK having to write the key twice, it's very readable IMO.

cbrnr:
Also, you are comparing float values using `np.testing.asert_equal()`. We'll see how this goes once the test file is available, but my assumption is that is might fail due to floating point precision. Therefore, it might be necessary to change these comparisons to `np.testing.assert_approx_equal()` or `np.testing.assert_almost_equal()`.

src/pyxdf/pyxdf.py
jamieforth: If you only have 1 sample, the duration is 0, so wouldn't you end up with `1/0` in that case?

In my view you can only calculate a duration with two or more samples, and `if len(stream.time_stamps) > 1` assures this.

Without `-1` the effective sample rate in `minimal` is calculated as 11.25 Hz but it should be 10 Hz, so I think the -1 is correct.

test/test_data.py
jamieforth: That was just how it was implemented in the original test, but yes it probably should be it's own separate test.

test/test_data.py
jamieforth: These parameters are used, they are passed to `load_xdf`.

```
test/test_data.py::test_minimal_file[False-False] PASSED           [ 12%]
test/test_data.py::test_minimal_file[False-True] PASSED            [ 25%]
test/test_data.py::test_minimal_file[True-False] PASSED            [ 37%]
test/test_data.py::test_minimal_file[True-True] PASSED             [ 50%]
test/test_data.py::test_empty_streams_file[False-False] PASSED     [ 62%]
test/test_data.py::test_empty_streams_file[False-True] PASSED      [ 75%]
test/test_data.py::test_empty_streams_file[True-False] PASSED      [ 87%]
test/test_data.py::test_empty_streams_file[True-True] PASSED       [100%]
```

Thanks for adding the missing `skipif`!

src/pyxdf/pyxdf.py
cbrnr: Yes, you are right!

test/test_data.py
cbrnr: What I meant is that in `test_empty_streams`, you are not testing for the different values of `dejitter_timestamps` and `synchronize_clocks`. All tests are completely independent of whether these are set to `True` or `False`.

test/test_data.py
cbrnr: Or we keep it in this test, because we already have a parameterization of `dejitter_timestamps`. However, we might have to adapt some tests, because this specific part seems to rely on the other two arguments set to these exact values.

If it's too complicated, then let's move it to a new test.

src/pyxdf/pyxdf.py
jamieforth: OK cool, I just push a change for the calculation of `effective_srate` when `dejitter_timestamps=True` as well.

This one was doing `len(time_stamps) / (duration + tdiff)`, which is fine when time-stamps are perfect, but off otherwise.

### Perfect clock

```python
nominal_srate = 1
t = np.arange(0, 5, nominal_srate)
(t[-1] - t[0] + nominal_srate) / len(t)
```

=> `1.0`  # Correct

```python
nominal_srate = 1
t = np.arange(0, 5, nominal_srate)
(t[-1] - t[0]) / (len(t) - 1)
```

=> `1.0`  # Correct


### Fast clock

```python
nominal_srate = 1
effective_srate = 1.1
t = np.arange(0, 5, effective_srate)
(t[-1] - t[0] + nominal_srate) / len(t)
```

=> `1.08`  # Incorrect

```python
nominal_srate = 1
effective_srate = 1.1
t = np.arange(0, 5, effective_srate)
(t[-1] - t[0]) / (len(t) - 1)
```

=> `1.1`  # Correct


### Slow clock

```python
nominal_srate = 1
effective_srate = 0.9
t = np.arange(0, 4, effective_srate)
(t[-1] - t[0] + nominal_srate) / len(t)
```

=> `0.919`  # Incorrect

```python
nominal_srate = 1
effective_srate = 0.9
t = np.arange(0, 4, effective_srate)
(t[-1] - t[0]) / (len(t) - 1)
```

=> `0.9`  # Correct

test/test_data.py
jamieforth: > All tests are completely independent of whether these are set to True or False.

The test is now conditional on synchronize_clocks.

I've kept the parametrization of dejitter even though it has minimal effect on idealised time-stamps - that feels like a reassuring thing to validate to me, but also happy to remove it if you prefer.

test/test_data.py
```diff
@@ -5,91 +5,390 @@
...
+        ]
+    )
+
+    np.testing.assert_equal(streams[i]["time_series"], s)
+    if synchronize_clocks:
+        np.testing.assert_almost_equal(streams[i]["time_stamps"], t, decimal=4)
+    else:
+        np.testing.assert_almost_equal(streams[i]["time_stamps"], t, decimal=8)
+
+    # Dejittering should have no effect because timestamps have minimal jitter.
```
cbrnr, line 387: I know this is super nitpicky, but I'd prefer comments to not end with a period. Can you remove them here (and elsewhere if present)?

jamieforth: All good I think this was the only one here, but there's quite a few in my other PRs but I'll fixup those before we move on.

test/test_data.py
jamieforth: I've separated this into a separate test, I think it makes more sense.

jamieforth:
I think I'm happy with this now unless you see any more issues?

PR #130, opened 2025-01-26
Handle negative time intervals when dejittering

jamieforth:
The core functionality change here is segmenting at negative time intervals when dejittering, following discussion in #127.

Minor changes:
- ~~Add `stream_id` as a property of `StreamData`. This isn't strictly necessary it just simplifies modularity having `stream_id` available within stream objects rather than passing it as a separate argument to functions of individual streams. It's not used extensively here, but appears more in #131 and #132~~.
- New generic `MockStreamData` class for testing, used here in `test_jitter_removal` and subsequent PRs.
- New function `_find_segment_indices`. Fairly simple function but the purpose is to formalise the mapping from a `bool` breakpoint array to segment indices. Refactoring allows it to be tested and reused in #131, where there appears to be a bug that is fixed by reusing this function.


jamieforth:
@cbrnr this is the next PR to review when you get chance!

cbrnr:
Sure, it might take me a couple of days though. I'd appreciate if @cboulay and/or @agricolab could also take a look so that at least two people have reviewed this PR. Thanks!

jamieforth:
@agricolab this PR follows [your suggestion](https://github.com/xdf-modules/pyxdf/pull/127#issuecomment-2527637009) in my previous PR. You were absolutely right about the complications of sorting and interaction with `_clock_sync`, so none of that is here. Proposals for those issues are in my follow-up PRs, but this one I think is relatively uncontroversial!

src/pyxdf/pyxdf.py
```diff
@@ -29,8 +29,8 @@
 class StreamData:
     """Temporary per-stream data."""
 
-    def __init__(self, xml):
-        """Init a new StreamData object from a stream header."""
+    def __init__(self, xml, stream_id):
+        """Init a new StreamData object from a stream header and ID."""
```
cbrnr, line 33: Is it really necessary to add a new parameter here? You only need the stream ID for logging, but you could extract it from the `xml` argument I think.

cbrnr: OK, thanks! Since that seems to be a private function, passing `stream_id` should be no problem at all, and it allows us to keep the `StreamData` class clean.

agricolab:
Looks good to me. Thanks for the effort!

agricolab:
P.S. Sorry for the delay. I got several notifications, but, whenever i clicked on the notification, it directed me to this: 

image
Had to click on `conversation` to see the actual comments - then the files changed worked again.

Anyways...

cbrnr:
Thanks, before I go ahead and merge, can you point me to a test that demonstrates that dejittering with positive intervals continues to work as before (i.e., passes on `main` and in this PR)? I assume all added tests here deal with negative time intervals, right?

test/test_detect_breaks.py
```diff
@@ -0,0 +1,141 @@
...
+    stream = MockStreamData(time_stamps=time_stamps, tdiff=1)
+    # if diff > 1 -> 1
+    b_breaks = _detect_breaks(stream, threshold_seconds=1, threshold_samples=0)
+    breaks = np.where(b_breaks)[0]
+    assert breaks.size == 1
+    assert breaks[0] == 0
+    # if diff > 0.1 -> 4
+    b_breaks = _detect_breaks(stream, threshold_seconds=0.1, threshold_samples=0)
+    breaks = np.where(b_breaks)[0]
+    assert breaks.size == len(time_stamps) - 1
```
jamieforth, line 88: Hi @cbrnr, the tests cover both positive and negative time intervals. This first batch of `_detect_breaks` tests all pass with the existing implementation.

jamieforth: And all the tests for `_jitter_removal` except `test_jitter_removal_two_segments_non_monotonic` and `test_jitter_removal_glitch` pass on `main`.

jamieforth:
> P.S. Sorry for the delay. I got several notifications, but, whenever i clicked on the notification, it directed me to this:

@agricolab that was probably my bad, I forced pushed while the PR was still in draft so that probably accounts for the weirdness, sorry about that.

test/test_detect_breaks.py
cbrnr: Thanks @jamieforth! We're squashing before merging anyway, so the commits don't matter at all.

test/mock_data_stream.py
```diff
@@ -0,0 +1,158 @@
...
+        *,
+        stream_id=1,
+        srate=None,
+        tdiff=None,
+        fmt="float64",
+        time_stamps=[],
+        time_series=None,
+        clock_times=[],
+        clock_values=[],
+    ):
```
cbrnr, line 17: Could you add a docstring and describe all parameters here? IMO these are not all immediately obvious (especially `tdiff`).

test/mock_data_stream.py
```diff
@@ -0,0 +1,158 @@
...
+        self.stream_id = stream_id
+        # Validate srate and tdiff if both provided
+        if srate is not None and tdiff is not None:
+            assert 1 / srate == tdiff
```
cbrnr, line 21: Maybe raise an exception here instead of just asserting?

test/test_detect_breaks.py
```diff
@@ -0,0 +1,141 @@
...
+    breaks = np.where(b_breaks)[0]
+    assert breaks.size == 1
+    assert breaks[0] == 0
+    # if diff < 0 or diff > 0.1 -> 7
+    b_breaks = _detect_breaks(stream, threshold_seconds=0.1, threshold_samples=0)
+    breaks = np.where(b_breaks)[0]
+    assert breaks.size == len(time_stamps) - 1
+
+
+def test_detect_breaks_non_monotonic_gap_in_positive():
```
cbrnr, line 124: Same comment on duplication here, this test is identical to the previous one except for `time_stamps`. Do you think it is more readable like this, or would parameterizing be an option? I'm not insisting on parameterization, I'm just curious what you think.

test/test_segment_indices.py
```diff
@@ -0,0 +1,77 @@
...
+            for n_breaks in range(0, size + 1)
+        ]
+        for breaks in [np.array(breaks, dtype=int) for breaks in breaks_list]
+    ],
+)
+def test_segments_all(size, n_breaks, breaks, expected):
+    b_breaks = np.full(size, False)
+    b_breaks[breaks] = True
+    segments = _find_segment_indices(b_breaks)[0]
+    assert segments == expected
```
cbrnr, line 77: Does this test contain all smaller previous tests?

test/mock_data_stream.py
jamieforth: Done!

test/mock_data_stream.py
jamieforth: Indeed raising an exception is better. I also added some more tests for the expected errors.

test/mock_data_stream.py
```diff
@@ -0,0 +1,158 @@
...
+    np.testing.assert_equal(
+        stream.time_stamps, [float(sample[0]) for sample in stream.time_series]
+    )
+    assert stream.nchans == 1
+    assert stream.clock_times == clock_times
+    assert stream.clock_values == clock_values
+
+
+@pytest.mark.parametrize("nchans", [1, 2])
+def test_mock_stream_timeseries_num(nchans):
```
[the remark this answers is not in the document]
jamieforth: Yeah parameterizing makes sense here, the duplication made it difficult to see exactly what was going on.

test/test_segment_indices.py
jamieforth: Yes it does, I was eering on the side of caution! I liked the exhaustive parameterization but it's arguably not the most intuitive block of code so I wanted to recompute the simple tests as a sanity check. I guess there's not need to keep that though so I've changed it to continue from where the small tests leave off.

I like the simple tests for readability, but now I think of it those are also more readable as a single parameterized test.

jamieforth:
@cbrnr I've added changelog entries for this and #129.

jamieforth:
@cbrnr @agricolab do you have any more thoughts about this PR? I've been using it a fair bit and haven't encountered any issues. If you think it's worth merging I'd say it's good to go. 

More selfishly I'd really appreciate your thoughts on #131 when you have some time :)

CHANGELOG.md
```diff
@@ -1,6 +1,16 @@
 ## [UNRELEASED] - YYYY-MM-DD
 ### Added
 - Add new `case_sensitive` parameter to `match_streaminfos`, defaulting to `False` to maintain previous behavior; when `False`, stream properties are matched more leniently ([#134](https://github.com/xdf-modules/pyxdf/pull/134) by [Stefan Appelhoff](https://github.com/sappelhoff))
+- Expand tests for dejitter segments to cover all initialisation contexts ([#129](https://github.com/xdf-modules/pyxdf/pull/129) by [Jamie Forth](https://github.com/jamieforth))
+- Generic `MockStreamData` class for unit testing ([#130](https://github.com/xdf-modules/pyxdf/pull/130) by [Jamie Forth](https://github.com/jamieforth))
+- New tests for jitter removal and segmentation ([#130](https://github.com/xdf-modules/pyxdf/pull/130) by [Jamie Forth](https://github.com/jamieforth))
```
cbrnr, line 6: I don't think we need to include additions/changes to our tests here (at least we've never done that before).

jamieforth: There was a previous entry about tests so I just followed that, but yes I don't think it's necessary to document tests here.

cbrnr:
Thanks a lot @jamieforth!

Issue #135, opened 2025-05-08
playback_lsl errors


cboulay:
Grouping a couple errors together so I only have to create 1 PR.

Problem 1: 

Since e8817c06bd8781b91d5d6a8e9054c6288dde3648, if the sleep duration (default 5 msec) is shorter than the time between two samples in any of the streams, `stop_idx` will not be larger than `start_idx` and `all_streams_exhausted` will remain `True`, causing the loop to exit and terminate playback prematurely.

I think the solution is for the streams-exhausted check to be based on (1) looping disabled and (2) the streams' time ranges. This will probably require some new state variables.

Problem 2:

From the playback entry point, one of the arguments to `LSLPlaybackClock` is `loop_time=wrap_dur if loop else None`, and this value gets assigned to `self._boundary`. However, the clock's `t0` property tries to do `self._n_loop * self._boundary` which doesn't work when `loop` is `False` and therefore `self._boundary` is `None`.

The easy solution is to fix the argument to be `loop_time=wrap_dur if loop else 0.0`, which is probably the correct solution (vs fixing `t0` property) because the `loop_time` annotation is non-optional `float`.

PR #136, opened 2025-05-08
135 playback lsl - Fix for errors when not looping or when sleep time is shorter than sample interval

cboulay:
This PR fixes the issues described in #135


cbrnr:
LGTM! I assume you have tested this and the changes fix the issues. Would it be possible to add test cases which fail on `main` but pass here?

cboulay:
Do we want to add pylsl as a test dependency?

cbrnr:
I would prefer to add it as an extra, because users would also need pylsl in order to use the playback CLI.

cbrnr:
Why can you not just use pylsl?

cboulay:
I added a test that simply attempts to playback one of the files already in the example-files folder. This test fails without the change to the loop argument in this PR.

I do not wish to test, however, the other problem this PR fixes which was the early termination when the time between samples was larger than the sleep time. This is rather difficult to test because it requires a new contrived test file with long time intervals and a separate thread to read the stream and verify the entire contents were passed through.

(FYI: Adding this test also prompted me to add a `noble` release to liblsl -- glad to have that done.)

cboulay:
> Why can you not just use pylsl?

pylsl does not come with the liblsl binaries on Linux and Mac because it assumes you'll have liblsl installed at the system level.

cbrnr:
Is there a reason why these packages do not bundle the binary libs?

cboulay:
Yes.
Previously we were shipping binaries inside the wheels for all platforms. However, the MacOS binaries were annoying to maintain because of security issues. The linux binaries were causing more problems than they were worth because the docker containers that were available to produce `manylinux` binaries were not producing binaries that were meeting the needs of all users and it was actually worse to provide non-working binaries than it was to provide no binaries with a helpful error message. Packaging tools were not able to produce bespoke limited-scope wheels that pypi would accept that would only be provided to platforms on which they worked without also feeding them to platforms on which they didn't work.

Is this going to be a problem for this PR?

cbrnr:
Thanks for clarifying. I think this is problematic, because I am hesitant to add a dependency (even if optional) which requires manual tinkering. Therefore, I'd rather not include the test and the pylsl dependency (and return to the status quo).

I'm happy to discuss pylsl packaging in a separate issue. I do have some limited experience with cibuildwheel, which I use to create binary wheels for SleepECG (which contains a C extension). Ideally, pylsl should provide self-contained binary wheels for all three platforms, since this has been the standard way to ship such things for quite some time (and this is also what users will expect). Maybe this is worth another look (tools might have matured since you last tried)?

cboulay:
There’s non-Debian, arch Linux, Android, all the different flavours of arm (pi), etc. These are all platforms where users complained that pylsl wasn’t working for them.
The thing about these users is they are typically able to follow the instructions in an error message to install missing liblsl, but they get stuck when they have an incompatible liblsl.so. 

Anyway, I’ll edit the PR to remove the tests when I have a moment.

cbrnr:
> There’s non-Debian, arch Linux, Android, all the different flavours of arm (pi), etc. These are all platforms where users complained that pylsl wasn’t working for them.
> The thing about these users is they are typically able to follow the instructions in an error message to install missing liblsl, but they get stuck when they have an incompatible liblsl.so.

Yeah, but this should not be a problem nowadays. Even ARM64 builds work out of the box for me.

.github/workflows/test.yml
```diff
@@ -8,6 +8,10 @@ on:
       - main
   workflow_dispatch:
 
+env:
+  LSL_RELEASE_URL: "https://github.com/sccn/liblsl/releases/download/v1.16.2"
+  LSL_RELEASE: "1.16.2"
+
```
cbrnr, line 14: This should be removed.

pyproject.toml
```diff
@@ -43,6 +43,11 @@ Repository = "https://github.com/xdf-modules/pyxdf"
 Issues = "https://github.com/xdf-modules/pyxdf/issues"
 Changelog = "https://github.com/xdf-modules/pyxdf/blob/main/CHANGELOG.md"
 
+[project.optional-dependencies]
+playback = [
+    "pylsl>=1.17.6",
+]
+
```
cbrnr, line 50: This should be removed.

cbrnr (commented):
You forgot to remove these two things, otherwise LGTM!

pyproject.toml
cboulay: Is this not still useful for people who want to do `pip install "pyxdf[playback]"`?
(with the caveat that it's only seamless for Windows users or users who already have liblsl on their system)

.github/workflows/test.yml
cboulay: Oops! Thanks. I did a force push to amend the previous 'remove playback test' commit.

pyproject.toml
cbrnr: Sure, we can keep it!

cbrnr:
Thanks @cboulay!

PR #140, opened 2025-09-24
detected breaks not only if diffs are positive, but use abs(diffs)

### LINKED ISSUE #137: What does "Segments and clock-segments differ" mean?
Hi!
I get the warning "Stream 3: Segments and clock-segments differ", coming from:
https://github.com/xdf-modules/pyxdf/blob/fc32bc21429179685fc2364654359262714a3692/src/pyxdf/pyxdf.py#L401C1-L401C82

Unfortunately, I don't know if this is something I shoudl take into account or not (because the recording was interrupted by the sender multiple times, I do expect different segments and possibly clock "resets").

Thanks for your input!
Best, Benedikt

behinger:
Fixes #137

I did not run tests locally because I dont have the tooling installed

### DIFF
--- a/src/pyxdf/pyxdf.py
+++ b/src/pyxdf/pyxdf.py
@@ -739,9 +739,8 @@ def _detect_breaks(stream, threshold_seconds=1.0, threshold_samples=500):
         b_breaks : 1D bool array representing breaks between values.
     """
     diffs = np.diff(stream.time_stamps)
-    b_breaks = (diffs < 0) | (
-        diffs > np.max((threshold_seconds, threshold_samples * stream.tdiff))
-    )
+
+    b_breaks = np.abs(diffs) > np.max((threshold_seconds, threshold_samples * stream.tdiff))
     return b_breaks
 
 


cbrnr:
@jamieforth can you take a look? @behinger if you fix the style errors we'll see if the tests succeed (or if they need to be adapted).

jamieforth:
I don't think I can push to this branch but there's updated tests here: behinger/pyxdf#1. All tests pass except for one: `test_jitter_removal_glitch` involving the calculation of effective sample-rate.

I still tend to think this is more of a hack for a broken LSL device than a bug in pyxdf, but `abs(diff)` is a nice practical solution (it was also my first thought too #127). However, it can cause subtle problems with parts of the code (or subsequent resampling, analysis etc.) that assume monotonicity. It's probably possible to adapt other code to not assume time-stamps are always strictly ordered, but it's a very convenient assumption to be able to rely upon.

The failing test is one example: it fails because the following timestamps are now segmented into three segments.
```
[1, 2, 3, 4] + [6, 5] + [7, 8, 9, 10]
```
The duration of the middle segment is -1, which leads to the incorrect calculation of effective sample rate.

We could change the test if we deem this to be the correct behaviour?

cbrnr:
I agree with @jamieforth that this change should probably not replace the current behavior if the issue is the device. I didn't follow the entire discussion, but would it not be better to write a small tool to fix the broken timestamps in the first place?

cbrnr:
I'm OK with doing what the MATLAB lib does. It still seems like it's probably an upstream issue, but let's just be consistent with MATLAB here.

Regarding the test, I'll leave it up to you, though I'd slightly prefer option 2. But it's your call!

behinger:
I have seen jitter > 1/fs on other computers with pylsl - maybe a function of buffersize? But in any case, keeping same as matlab is good

jamieforth:
Cool, xfail for the test was my instinct too.

@behinger I think you'll need to merge my PR in your fork to get the updated tests here, unless @cbrnr has a smarter suggestion.

behinger:
argh super sorry - I missed this somehow. Merged now!

PR #169, opened 2026-03-31
Fix TypeError when channel/acquisition metadata contains None values

sappelhoff:
Disclaimer: I created this PR and the content using Claude.

---

## Summary

`playback_lsl.py` crashes with a `TypeError` when an XDF file contains channel or acquisition metadata fields that are present but have `None` values:

```
TypeError: descriptor 'encode' for 'str' objects doesn't apply to a 'NoneType' object
```

This happens in `pylsl`'s `append_child_value`, which calls `str.encode()` on the value passed to it.

## Fix

- Skip metadata fields whose value is `None` (guard added to both the `acquisition` and `channel` loops)
- Wrap values in `str()` so non-string scalars (e.g. a numeric `precision` field) are also handled correctly

## Test plan

- [x] Reproduce with an XDF file that has `None` channel metadata (e.g. a missing `unit` field) — previously crashed, now streams successfully
- [x] Verify that XDF files with fully populated metadata still stream correctly

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,6 @@
 ## [UNRELEASED] - YYYY-MM-DD
+### Fixed
+- Fix `playback_lsl.py` for cases where no metadata is available in the to-be-streamed XDF file ([#160] by [Stefan Appelhoff](https://github.com/sappelhoff))
 
 ## [1.17.3] - 2026-01-20
 ### Fixed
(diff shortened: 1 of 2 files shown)


sappelhoff:
cc @cbrnr

cbrnr:
Thanks! LGTM! Can you add a changelog entry please?

cbrnr:
Thanks @sappelhoff!

PR #170, opened 2026-06-14
Precompute little-endian sample dtype once per stream

sappelhoff:
**NOTE** I generated this PR with the help of Claude as part of a general memory/speed profiling I am doing for a private package of mine.

## What

`_read_chunk3` reinterprets each numeric sample buffer as little-endian via `np.frombuffer(raw, dtype=s.dtype.newbyteorder("<"), count=s.nchns)`. `s.dtype.newbyteorder("<")` returns the same dtype object every time but was being recomputed **once per sample**. This moves it to `StreamData.__init__` (`s.dtype_le`, alongside the existing `dtype`/`samplebytes` pre-calcs) and reuses it.

## Why it's behaviour-preserving

- XDF sample data is little-endian, and the original code already hardcoded `"<"` regardless of platform — `s.dtype_le = s.dtype.newbyteorder("<")` is the identical value, so output is **byte-identical** (including on big-endian hosts).
- `dtype_le` is set under the same `if self.fmt != "string":` guard that already gates `self.dtype`, and its only use is the numeric branch of `_read_chunk3`. String streams never set or read it. Empty streams never call `_read_chunk3`.
- `newbyteorder` is dtype-generic (no type-specific branch), so all numeric formats behave identically.

The `.pxd` Cython declaration is updated to keep `StreamData` in sync.

## Verification

- Full suite: **372 passed, 1 xfailed** (with `example-files/` cloned), covering int16/int32/float32/double64/string/empty streams — every sample-reading path.
- **Byte-identical** per-stream `time_series`/`time_stamps` md5 on a real 224 MB recording (99-ch EEG @ 500 Hz + marker/logging/notes/misc streams), baseline vs this branch.
- ~**8% faster** `load_xdf` on that recording (median 1.014 s → 0.935 s, 6 reps), as the per-sample `newbyteorder` (553k calls) is eliminated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## [UNRELEASED] - YYYY-MM-DD
 ### Changed
-- Precompute the little-endian sample dtype once per stream instead of per sample in `_read_chunk3`, speeding up loading of large numeric streams (~8% faster on a 553k-sample, 99-channel EEG recording) with byte-identical output ([#PR] by [Stefan Appelhoff](https://github.com/sappelhoff))
+- Precompute the little-endian sample dtype once per stream instead of per sample in `_read_chunk3`, speeding up loading of large numeric streams (~8% faster on a 553k-sample, 99-channel EEG recording) with byte-identical output ([#170](https://github.com/xdf-modules/pyxdf/pull/170) by [Stefan Appelhoff](https://github.com/sappelhoff))
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


cbrnr:
Very nice, thanks @sappelhoff!

sappelhoff:
Sure! Could we perhaps get a release with this? 🙏 the speedup is modest but noticeable :)

cbrnr:
Yes, I will try to do it today or tomorrow at the latest.

